PR #47, opened 2020-09-17
fix build for libfranka

jethrokuan:
`libfranka` requires `zlib` and `pcre` to build.

In addition, I set `-DBUILD-TESTS=OFF` for libfranka, because it attempts to download `googletest`, and for some reason it fails to resolve the fetch from GitHub.


lopsided98:
I have opened a PR in nixpkgs to fix poco (https://github.com/NixOS/nixpkgs/pull/98186), and added it to my nixpkgs fork for now. I also disabled `libfranka`'s tests.

`libfranka` and its dependents all build successfully now.

PR #52, opened 2020-10-08
Fix: flake with devShell

GTrunSec:
add devShell to flake

test output:
```
 nix develop -c rostest --help
Usage: rostest [options] [package] <filename>

Options:
  -h, --help            show this help message and exit
  -t, --text            Run with stdout output instead of XML output
  --pkgdir=PKG_DIR      package dir
  --package=PACKAGE     package
  --results-filename=RESULTS_FILENAME
                        results_filename
  --results-base-dir=RESULTS_BASE_DIR
                        The base directory of the test results. The test
                        result file is created in a subfolder name PKG_DIR.
  -r, --reuse-master    Connect to an existing ROS master instead of spawning
                        a new ROS master on a custom port
  -c, --clear           Clear all parameters when connecting to an existing
                        ROS master (only works with --reuse-master)

```
 TODO.
-  set `defaultPackage. ` for nix build
-  add CI  for Nix flake check such as `nix develop -c echo ok  and nix build nix-run`

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/examples/turtlebot3-gazebo-shell.nix
+++ b/examples/turtlebot3-gazebo-shell.nix
@@ -1,0 +1,26 @@
+# Run:
+# roslaunch turtlebot3_gazebo turtlebot3_world.launch
+# roslaunch turtlebot3_teleop turtlebot3_teleop_key.launch
+{ pkgs,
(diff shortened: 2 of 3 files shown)


lopsided98:
I'm not sure it makes sense for this flake to have a `devShell` output. Every user will need different packages, so the turtlebot example shell won't be useful to very many people and might just be confusing.

PR #69, opened 2021-01-07
Bump pcl-ros to 1.7.2 for PCL 1.11 support

3noch:
I doubt this is done in a way that you actually want to merge but I wouldn't mind learning what the best way would be :D


3noch:
At the very least this should also bump `pcl_conversions` since they're paired in the same repo.

lopsided98:
I backported another upstream patch, which fixed the build. The fix is currently in the staging branch, but I will merge it to master once I fix a few other issues.

3noch:
Sounds good. Thanks.

PR #98, opened 2021-05-10
gazebo: Fix build with allowAliases = false

jtojnar:
- Do not use pkg-config alias
- gazebo: Avoid aliases


jtojnar:
Note that other packages still use the libusb alias. Maybe superflore needs updating.

lopsided98:
https://github.com/ros/rosdistro/pull/29496

PR #99, opened 2021-05-10
melodic/franka-ros: add missing dependencies

fricklerhandwerk:
Unfortunately this does not build because of `defusedxml`. For the project where the patches were needed I used `nixpkgs-20.09`. There `defusedxml` is still at 0.6.0 and compatible with Python2.7.

I don't get why `overridePythonAttrs` does not apply properly. Can anyone help out?

```log
while evaluating the attribute 'defusedxml' at /home/vg/src/nix-ros-overlay/pkgs/default.nix:45:7:
while evaluating the attribute 'defusedxml.overridePythonAttrs' at /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/pkgs/top-level/python-packages.nix:1788:3:
while evaluating 'callPackageWith' at /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/lib/customisation.nix:117:35, called from /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/pkgs/top-level/python-packages.nix:1788:16:
while evaluating 'makeOverridable' at /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/lib/customisation.nix:67:24, called from /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/lib/customisation.nix:121:8:
while evaluating anonymous function at /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/pkgs/development/python-modules/defusedxml/default.nix:1:1, called from /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/lib/customisation.nix:69:16:
while evaluating 'makeOverridablePythonPackage' at /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/pkgs/top-level/python-packages.nix:32:37, called from /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/pkgs/development/python-modules/defusedxml/default.nix:8:1:
while evaluating 'makeOverridable' at /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/lib/customisation.nix:67:24, called from /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/pkgs/top-level/python-packages.nix:34:12:
while evaluating anonymous function at /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/pkgs/development/interpreters/python/mk-python-derivation.nix:30:1, called from /nix/store/dxy4zxqgs0jdf4zswj0j9frbqy21qli7-source/lib/customisation.nix:69:16:
defusedxml-0.7.0 not supported for interpreter python2.7
```

Tested with

```nix
with (import ./. {});
mkShell {
  nativeBuildInputs = [
    (with rosPackages.melodic; buildEnv {
      paths = [
        desktop-full
        franka-ros
      ];
    })
  ];
}
```

### DIFF
--- a/distros/melodic/overrides.nix
+++ b/distros/melodic/overrides.nix
@@ -16,6 +16,29 @@ rosSelf: rosSuper: with rosSelf.lib; {
 
   gazebo = self.gazebo_9;
 
+  image-view = rosSuper.image-view.overrideAttrs (old: with self; {
+    # the problem is very similar to this:
+    # https://github.com/ros/ros-overlay/issues/607
+    cmakeFlags = [
+      "-DGTK2_GLIBCONFIG_INCLUDE_DIR:PATH=${glib.out}/lib/glib-2.0/include"
+      "-DGTK2_GDKCONFIG_INCLUDE_DIR:PATH=${gtk2.out}/lib/gtk-2.0/include"
+    ];
+  });
+
+  libfranka = rosSuper.libfranka.overrideAttrs (old: {
+    propagatedBuildInputs = with self; [ zlib pcre ] ++ old.propagatedBuildInputs;
+  });
+
+  moveit-core = (rosSelf.callPackage ../noetic/moveit-core {}).overrideAttrs (old: {
+    pname = "ros-melodic-moveit-core";
+    src = fetchTarball {
+      url = "https://github.com/ros-gbp/moveit-release/archive/refs/tags/release/melodic/moveit_core/1.0.7-1.tar.gz";
+      sha256 = "1b41dwdfwpywkf706403pk5ak810ijr46gri0phhi0wbnr071mqn";
+    };
+  });
+
+  moveit-ros-perception = (rosSelf.callPackage ./moveit-ros-perception {});
+
   pcl-ros = rosSuper.pcl-ros.overrideAttrs ({
     patches ? [], ...
   }: {


lopsided98:
Python 2 is EOL, and nixpkgs is dropping support for it. I don't really want to start maintaining workarounds, especially when ROS noetic (and melodic, for the most part) supports Python 3. IMO, your best course of action is to figure out how to move to Python 3.

Regarding `fcl`, you are not the first person who has asked for it, but I would prefer it to be in nixpkgs. Previously, I have been reluctant to take on the maintenance responsibilities myself because I don't use the package, but I decided to go ahead and create a PR to add it: https://github.com/NixOS/nixpkgs/pull/123060. I would appreciate it if you would consider helping adding yourself as a maintainer to fcl and its dependencies, and helping with maintenance in the future.

lopsided98:
One thing to note that is slightly confusing here is that fcl is a ROS package in ROS 1, but a system dependency in ROS 2. So in your case, you only need libccd in nixpkgs, but newer ROS versions need fcl as well.

fricklerhandwerk:
> Python 2 is EOL, and nixpkgs is dropping support for it. I don't really want to start maintaining workarounds, especially when ROS noetic (and melodic, for the most part) supports Python 3. IMO, your best course of action is to figure out how to move to Python 3.


That is all true, and yet with `nix` and `git` we finally have the possibility to keep legacy software operational indefinitely, without much hassle, for people who can't "just migrate" to newer versions because of their own legacy code or compatibility, or for whatever reason.

What speaks against having branches that build against `nixpkgs` releases, like `home-manager` for example? No need to keep any of them in Cachix.

Although my question here was more about how to compose Python overrides in the first place, this has nothing to do with Python 2 support in particular. It may be required to patch Python packages at some point anyway. One can't always wait for upstream, and sometimes it doesn't even make sense to upstream at all.

lopsided98:
Nix certainly makes it possible to keep a given setup working indefinitely, but that doesn't necessarily mean it makes sense to support arbitrary combinations of new and old software. In my case, I have decided to support `nixos-unstable`, because that is most useful to me. I work to make this repo useful to others on a best effort basis, but I maintain it in my spare time, so I have to make choices about what configurations to support.

To address your real question: you are running into the following nixpkgs issue: https://github.com/NixOS/nixpkgs/issues/48663

fricklerhandwerk:
Alright, agreed.

Updated against current `nixpkgs` that already has your `fcl` package and migrated the fixes to `melodicPython3`. But there is an entirely new build failure where `ompl` has an impure dependency on `pybind11`. It's probably better now than it was before, so whoever wants to use `franka-ros` will (hopefully) only have to patch that. Would you still merge that?

PS: I've already spent quite a bit of time to make this run on `nixpkgs-20.09` and regular `melodic`, and just wanted to contribute back. But it's quite exhausting to play catch-up with moving targets. That's the main reason why I'd prefer sticking to a release and at least make large chunks build out of the box. Would be much easier to accumulate fixes, as otherwise they simply get lost once those random breakages get introduced.

lopsided98:
`melodicPython3.franka-ros` now builds on `master`. I'm not sure what issue you were running into with OMPL.

I don't think the breakage situation is as bad as you're making it out to be. `franka-ros` has never worked in the past because I never put in the effort to fix the various issues (many of which don't have anything to do with a particular nixpkgs branch). I monitor for new build failures, so now that it builds it should continue to do so and I will fix any issues that come up. My perspective is that I know that I will have to fix these breakages someday, so I might as well deal with them as they arise, rather than all at once when I update to a new stable branch.

fricklerhandwerk:
Ah I see, you recently fixed all the issues, including the ones from this patch, before I even noticed. In that case, never mind.

Issue #46, opened 2020-09-16
Practical usage example


jethrokuan:
Hi, thank you for attempting to package ROS with Nix. I'm struggling a little to wrap my head around how to use this in an actual ROS catkin workspace. I've looked at the example, and it does a local import of the `default.nix` and `overlays.nix` file.

My experience with Nix is limited, but the usage of the Emacs overlay is simple:

```
{
  nixpkgs.overlays = [
    (import (builtins.fetchTarball {
      url = https://github.com/nix-community/emacs-overlay/archive/master.tar.gz;
    }))
  ];
}
```

I was hoping that the usage of the this overlay would be similar, which would then expose `rosPackages` under `pkgs`, but I couldn't get this to work. Do you have a sample catkin workspace using this overlay that I could look at? Do you also develop work on ROS packages and build them with Nix?

lopsided98:
You should be able to do something like this (not tested):
```
{
  nixpkgs.overlays = [
    (import (builtins.fetchTarball {
      url = "https://github.com/lopsided98/nix-ros-overlay/archive/master.tar.gz";
    }) + "/overlay.nix")
  ];
}
```

`default.nix` provides a pinned and sometimes patched (while I am waiting for a PR to be merged) version of nixpkgs with the ROS overlay applied. The packages that are uploaded to Cachix are built against this nixpkgs version, so you should use it if you don't want to build things yourself.

I only test against unstable, so you might run into problems if you use a stable version of nixpkgs.

My normal workflow is to create a `shell.nix` that imports this repo:
```
# You could use fetchTarball
{ pkgs ? import ../../../../Projects/ros/nix-ros-overlay { } }:
with pkgs;
with rosPackages.noetic;

mkShell {
  name = "ros-example";

  buildInputs = [
    (buildEnv { paths = [
      catkin
      rosbash
      roslaunch
      rospy
      rosnode
      mavros
      mavros-extras
      angles
      flann
      eigen
      geos
      tf2-geometry-msgs
      tf2-tools
      robot-state-publisher
      actionlib-tools

      mapviz
      mapviz-plugins
      tile-map

      rqt-reconfigure
    ]; })
  ];

  shellHook = ''
    export ROS_HOSTNAME=localhost
    export ROS_MASTER_URI=http://localhost:11311/
  '';
}
```
[...]

Issue #45, opened 2020-08-20
ROS2 example


lelongg:
Could someone provide an example on how to use this overlay with ROS2 please ?

lopsided98:
ROS2 support is pretty rough right now, mostly because I don't actually use it. I just tried a simple shell for foxy:
```
with import ../. {};
with rosPackages.foxy;

mkShell {
  buildInputs = [
    ros2topic
    ros2node
  ];
}
```
but was unable to get even `ros2 topic list` to work. It gives the following error:
```
>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'type support not from this implementation, at /build/rmw_fastrtps-release-release-foxy-rmw_fastrtps_cpp-1.2.0-1/src/publisher.cpp:75, at /build/rcl-release-release-foxy-rcl-1.1.7-1/src/rcl/node.c:276'

with this new error message:

  'rcl node's rmw handle is invalid, at /build/rcl-release-release-foxy-rcl-1.1.7-1/src/rcl/node.c:428'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
[ERROR] [1600297455.820599941] [rcl]: Failed to fini publisher for node: 1
Unknown error creating node: rcl node's rmw handle is invalid, at /build/rcl-release-release-foxy-rcl-1.1.7-1/src/rcl/node.c:428
```

I don't know enough about ROS2 to fix this easily, but I think it might be due to a missing dependency somewhere. `eloquent` seems to work though, so maybe try that?

What specific problem did you run into?

alexhenning:
Adding ` RMW_IMPLEMENTATION = "rmw_fastrtps_dynamic_cpp";` to the shell and `rmw-fastrtps-dynamic-cpp` to the list of packages seems to resolve the issue for me.

From https://github.com/ros2/rmw_fastrtps#two-different-rmw-implementations
> rmw_fastrtps actually provides not one but two different ROS 2 middleware implementations, both of them using Fast DDS as middleware layer: rmw_fastrtps_cpp and rmw_fastrtps_dynamic_cpp (note that directory rmw_fastrtps_shared_cpp just contains the code that the two implementations share, and does not constitute a layer on its own).

[...]

PR #100, opened 2021-05-13
README: add instructions for local overrides

fricklerhandwerk:
It took me some time to get going with my setup, and finding such instructions quickly would have helped me.

inspiration:
https://github.com/lopsided98/nix-ros-overlay/issues/46#issuecomment-696255968
https://github.com/lopsided98/nix-ros-overlay/issues/45#issuecomment-790617118

PS: Awesome project! Makes that monster of a system quite convenient to handle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,8 @@ Want to use ROS, but don't want to run Ubuntu? This project uses the power of [N
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
@lopsided98 anything speaking against merging this?

PR #137, opened 2021-09-27
Avoid hex literals in function code

acowley:
This is similar to the problem discussed in https://github.com/NixOS/nix/issues/5262


acowley:
I've been using this workaround for a week or so, and just today found that it's hitting more things in the nix ecosystem. The linked issue describes things fairly well.

PR #147, opened 2021-11-12
Use Python 3.8

### LINKED ISSUE #143: rqt_graph error
Trying to run `rqt_graph` from noetic results in,

```
$ rqt_graph
CompositePluginProvider.discover() could not discover plugins from provider "<class 'qt_gui.recursive_plugin_provider.RecursivePluginProvider'>":
Traceback (most recent call last):
  File "/nix/store/z4k8gh0r87h4015dghp34pv7c6lf2p9r-ros-noetic-qt-gui-0.4.2-r1/lib/python3.9/site-packages/qt_gui/composite_plugin_provider.py", line 57, in discover
    plugin_descriptors = plugin_provider.discover(discovery_data)
  File "/nix/store/z4k8gh0r87h4015dghp34pv7c6lf2p9r-ros-noetic-qt-gui-0.4.2-r1/lib/python3.9/site-packages/qt_gui/recursive_plugin_provider.py", line 53, in discover
    plugin_descriptors = self._plugin_provider.discover(discovery_data)
  File "/nix/store/dxsk9ddd6asj7bv5vaf72b7z5jyin3f7-ros-noetic-rqt-gui-0.5.2-r1/lib/python3.9/site-packages/rqt_gui/ros_plugin_provider.py", line 67, in discover
    plugin_descriptors += self._parse_plugin_xml(package_name, plugin_xml)
  File "/nix/store/dxsk9ddd6asj7bv5vaf72b7z5jyin3f7-ros-noetic-rqt-gui-0.5.2-r1/lib/python3.9/site-packages/rqt_gui/ros_plugin_provider.py", line 128, in _parse_plugin_xml
    for library_el in root.getiterator('library'):
AttributeError: 'ElementTree' object has no attribute 'getiterator'
```

[This](https://stackoverflow.com/a/65138248/277078) suggests it's an incompatibility with python 3.9. So I rebuilt everything with `python38`, and `rqt_graph` now works.

I don't know if the way I did this is the right way, so I'm open to advice before opening a PR. What I did was set `python = python38;` in the `noetic` expression in `distros/default.nix`, and then `python3 = pythonOverridesFor self.python38;` in the `base` expression in `distros/distro-overlay.nix`. I do not know if the latter change was needed, but I can't test it right now as rebuilding everything takes quite a while.

If this looks right, I'll open a PR to make this change.

acowley:
Fixes #143

### DIFF
--- a/distros/default.nix
+++ b/distros/default.nix
@@ -16,7 +16,7 @@ self: super: {
 
     noetic = import ./distro-overlay.nix {
       distro = "noetic";
-      python = self.python3;
+      python = self.python38;
     } self super;
 
     dashing = import ./distro-overlay.nix {
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -25,13 +25,29 @@ let
     python = pythonOverridesFor python;
     pythonPackages = rosSelf.python.pkgs;
 
-    python3 = pythonOverridesFor self.python3;
+    python3 = pythonOverridesFor self.python38;
     python3Packages = rosSelf.python3.pkgs;
 
-    boost = self.boost.override {
+    boost = (self.boost.override {
       python = rosSelf.python;
       enablePython = true;
-    };
+      enableNumpy = true;
+    }).overrideAttrs (old: {
+      # boost python detection from mrt is somewhat fragile. If we
+      # build with python-3.8, for example, we get libraries with
+      # names like `libboost_python38.so`. mrt looks for
+      # `libboost_python3.so`, so we help it find the libraries it's
+      # looking for.
+      postFixup =
+        let suffix = super.lib.strings.concatStrings
+          (super.lib.splitVersion rosSelf.python.pythonVersion);
+        in old.postFixup or "" + ''
+        if [ ! -f $out/lib/libboost_python3.so ]; then
+          ln -s $out/lib/libboost_numpy${suffix}.so $out/lib/libboost_numpy3.so
+          ln -s $out/lib/libboost_python${suffix}.so $out/lib/libboost_python3.so
+        fi
+      '';
+    });
   };
 
   overrides = rosSelf: rosSuper: with rosSelf.lib; {


acowley:
I've added a related fix. The symptom here was that `lanelet2` could not build. It turned out to be `lanelet2-python` whose build failed because it could not find boost-python. This is because lanelet2 is using MRT that looks for `libboost_python3.so`, while we had `libboost_python38.so`. The change here adds a symlink if needed. I also enabled `numpy` in `boost`, as this is wanted by lanelet2.

lopsided98:
The Boost Python problem is not limited to mrt_cmake_modules. For several other packages, I use the `patchBoostPython` function to add the necessary version suffix in the CMake script. The problem originates from the fact that Debian and Ubuntu used to add `libboost_python.so` and `libboost_python3.so` symlinks that are not part of the standard Boost installation used on other distros. Since most ROS packages only target Ubuntu, they often use this non-standard suffix. Debian and Ubuntu stopped doing this with Boost 1.71, but it [seems like mrt_cmake_modules somehow still works on these distros](https://github.com/KIT-MRT/mrt_cmake_modules/issues/25) on Ubuntu but not Nix. On the other hand, as of Boost 1.73, the non-standard CMake components have been added upstream: https://github.com/boostorg/boost_install/issues/31.

nixpkgs currently defaults to Boost 1.69, but if we use 1.73 in this overlay these problems should go away. I can confirm that it fixes lanelet2. This newer version causes a few build failures (mostly in melodic) due to the removal of deprecated features, but these failures are relatively easy to fix.

Lastly, what package requires Boost Numpy? lanelet2 seems to build fine without it.

acowley:
I think the boost bumpy component came up while building `lanelet2-python`, probably from [here](https://github.com/KIT-MRT/mrt_cmake_modules/blob/master/cmake/Modules/FindBoostPython.cmake#L26). I'll give the boost update a try next week, thanks for explaining the situation!

PR #172, opened 2022-03-18
Add recurseForDerivations to rosPackages and distros

kfajdsl:
This is to make `nix search` work


lopsided98:
Done in bb7b9e3723d3148f5ba8db85f9e10c4475efdaff

PR #192, opened 2022-07-09
Fix cyclonedds build for darwin

tanshihaj:
Currently it is impossible to build https://github.com/lopsided98/nix-ros-overlay/blob/master/distros/foxy/cyclonedds/default.nix since it is CMake project which is configured with `-DCMAKE_SKIP_BUILD_RPATH=ON`. It breaks this calls: https://github.com/ros2-gbp/cyclonedds-release/blob/release/foxy/cyclonedds/docs/CMakeLists.txt#L22-L25 (This issue describes why it happens: https://github.com/NixOS/nixpkgs/issues/22060) and thus package build at least on darwin with new CMake.

Suggested changes fixes build for darwin and maybe for linux (didn't test it).

### DIFF
--- a/distros/foxy/cyclonedds/default.nix
+++ b/distros/foxy/cyclonedds/default.nix
@@ -2,7 +2,7 @@
 # Copyright 2022 Open Source Robotics Foundation
 # Distributed under the terms of the BSD license
 
-{ lib, buildRosPackage, fetchurl, cmake, cunit, maven, openjdk, openssl }:
+{ lib, stdenv, buildRosPackage, fetchurl, cmake, cunit, maven, openjdk, openssl }:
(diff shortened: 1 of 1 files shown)


lopsided98:
`-DCMAKE_SKIP_BUILD_RPATH=ON` isn't set anymore since https://github.com/NixOS/nixpkgs/commit/f7414e2b3ac69f220e19b1ef9ddfe8d64650c5ca, so this will become unnecessary soon, right?

Also, cyclonedds already builds fine on Linux, so it is not necessary to set `LD_LIBRARY_PATH`.

tanshihaj:
Yes, it seems that this PR now is pointless

PR #201, opened 2022-08-07
Fix humble packages for darwin platform and add turtlesim example

tanshihaj:
Trying to use this overlay in darwin, and found some missing packages for `humble` distro: `ros2pkg`, `ros2cli`, `ros2run` (but it seems that I added them wrong - according to README.md they autogenerated, but I didn't understand how to add this packages to autogenerator).

Also I've added most simple turtlesim example so people can quickly check if everything works fine.

### DIFF
--- a/distros/humble/generated.nix
+++ b/distros/humble/generated.nix
@@ -1200,6 +1200,8 @@ self: super: {
 
  ros2bag = self.callPackage ./ros2bag {};
 
+ ros2cli = self.callPackage ./ros2cli {};
+
  ros2cli-common-extensions = self.callPackage ./ros2cli-common-extensions {};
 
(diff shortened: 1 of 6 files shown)


lopsided98:
They are not auto-generated because they cannot resolve their dependency on `python3-pytest-timeout`. This package exists in nixpkgs, so it should just be a matter of adding it to rosdep.

tanshihaj:
Where can I find rosdep?

tanshihaj:
https://github.com/ros/rosdistro/pull/34130#pullrequestreview-1072704836 - I've updated rosdep. I only need to make rosdistro sync like in this PR https://github.com/lopsided98/nix-ros-overlay/pull/202, right?

lopsided98:
That runs automatically every Friday, but I just triggered a manual run.

PR #213, opened 2022-10-03
Update renamed Python dependencies

kknives:
Packages `pytestcov` and `pytestrunner` have been renamed in nixpkgs to `pytest-cov` and `pytest-runner`.
When using a package that depends on them, the following error output is displayed:
```
error: 'pytestrunner' has been renamed to/replaced by 'pytest-runner'
```
```
error: 'pytestrunner' has been renamed to/replaced by 'pytest-runner'
```
---
This PR renames all references (that I could find) to their updated name.

Thanks for all your work on this overlay, it really is amazing.

### DIFF
--- a/distros/foxy/tracetools-test/default.nix
+++ b/distros/foxy/tracetools-test/default.nix
@@ -15,7 +15,7 @@ buildRosPackage {
 
   buildType = "ament_cmake";
   buildInputs = [ ament-cmake pkg-config ];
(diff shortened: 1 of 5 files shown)


lopsided98:
Thanks! The colcon changes look fine, but the files in `distros/` are auto-generated; the dependency needs to be fixed in rosdistro: https://github.com/ros/rosdistro/blob/34f18b7a250c262563b3e8f2543b634dcb1d82b5/rosdep/python.yaml#L8439

kknives:
So, that needs to be fixed upstream. I can issue a new PR with just the colcon changes, if that's fine

kknives:
Great, one more thing, I can't seem to build colcon. I'll open an issue for that.

Issue #227, opened 2022-11-22
rviz does not build in noetic and melodicPython3


lelongg:
rviz fails to build with the following error

```
-- Found PythonInterp: /nix/store/xcaaly5shfy227ffs8nipxrd49b56iqq-python3-3.10.8/bin/python3 (found suitable version "3.10.8", minimum required is "3.10") 
CMake Warning at /nix/store/9fcvq10d1i5xy2v5gkhjyycv322w94j2-ros-noetic-python-qt-binding-0.4.4-r1/share/python_qt_binding/cmake/sip_helper.cmake:28 (message):
  SIP binding generator NOT available.
Call Stack (most recent call first):
  src/python_bindings/sip/CMakeLists.txt:42 (include)


-- Python binding generators: 
CMake Error at src/python_bindings/CMakeLists.txt:9 (message):
  No Python binding generator found.


-- Configuring incomplete, errors occurred!
``` 

reproducible with `nix build github:lopsided98/nix-ros-overlay#noetic.rviz `

lopsided98:
There's not much I can do about this. The latest versions of Qt5 are no longer compatible with SIP 4, so I switched to Shiboken/PySide2 (see https://github.com/lopsided98/nix-ros-overlay/issues/224). This fixed rqt, but rviz is not compatible with PySide2, see: https://github.com/ros-visualization/rviz/pull/1526

natto1784:
Thanks for maintaining all of the distros! Do you suggest anything that might help with building rviz? Maybe overriding qt5.qtbase?

natto1784:
Some issues you can follow: https://github.com/ros-visualization/rviz/issues/1777 https://github.com/ros-visualization/python_qt_binding/issues/121

kjeremy:
I'm running into this too. Maybe we can pin rviz to an older version from nix-ros-overlay?

natto1784:
Using [this patch](https://aur.archlinux.org/cgit/aur.git/tree/restore-sip4-support.patch?h=python-pyqt5-sip4) along with the old python-qt-binding override works with SIP4. I tested it as a global overlay for all distros and rviz compiled! Let's see if we can use this here for noetic and melodic specifically while using shiboken/pyside2 for other distros.

PR #245, opened 2023-02-07
distros/ros1-overlay: fix rviz build

natto1784:
By making sip-4 compatible with pyqt5 which is required by python-qt-binding, a dependency of rviz, using this patch:

https://aur.archlinux.org/cgit/aur.git/plain/restore-sip4-support.patch?h=python-pyqt5-sip4&id=6e712e6c588d550a1a6f83c1b37c2c9135aae6ba

Should fix https://github.com/lopsided98/nix-ros-overlay/issues/227

Signed-off-by: Amneesh Singh <natto@weirdnatto.in>


natto1784:
@lopsided98 can you take a look at this

PR #250, opened 2023-03-13
rmw-cyclonedds-cpp: Fix < 1.4.1 building with GCC 12.1

hacker1024:


lopsided98:
I've already applied this patch in foxy, and it is included in 1.3.4 in humble, so I don't think this is necessary.

PR #251, opened 2023-03-13
spacenav: Add required X11 dependency

hacker1024:


lopsided98:
Sorry for not getting to this sooner. This should be fixed in nixpkgs; `libspnav` needs to propagate `libX11`.

PR #253, opened 2023-03-21
fix: add zeromqt to behaviortree-cpp-v3

mbeutelspacher:
When I try to link against behaviortree-cpp-v3 I get undefined reference errors otherwise even if zeromqt is present in my env

### DIFF
--- a/distros/ros1-overlay.nix
+++ b/distros/ros1-overlay.nix
@@ -1,6 +1,11 @@
 self:
 rosSelf: rosSuper: with rosSelf.lib; {
 
+  behaviortree-cpp-v3 = rosSuper.behaviortree-cpp-v3.overrideAttrs
+    ({ propagatedBuildInputs ? [ ], ... }: {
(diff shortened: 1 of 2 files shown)


wentasah:
I don't think this is necessary. The [PR to nixpkgs](https://github.com/NixOS/nixpkgs/pull/224099), which should fix this was just merged. Therefore, once `github:lopsided98/nixpkgs/nix-ros` gets updated, the problem with behaviortree-cpp-v3 (and perhaps also with other packages) will disappear.

movefasta:
if overriding the same in ros1 and ros2 cases better to update [distros/distro-overlay.nix](https://github.com/lopsided98/nix-ros-overlay/blob/master/distros/distro-overlay.nix), that is common for both versions:
https://github.com/lopsided98/nix-ros-overlay/blob/master/distros/distro-overlay.nix#L426-L427

lopsided98:
Can you check whether this is still a problem, and, if so, provide an example package that fails?

mbeutelspacher:
No problem anymore ️❤️

Issue #248, opened 2023-02-17
`humble.rviz2` doesn't work


koalp:
I try to run rviz2 but after solving a first issue with qt I run in the following issue:

```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  failed to initialize rcl init options: failed to get symbol 'rmw_init_options_init' due to Environment variable 'AMENT_PREFIX_PATH' is not set or empty, at /build/rmw_implementation-release-release-humble-rmw_implementation-2.8.1-2/src/functions.cpp:171, at /build/rcl-release-release-humble-rcl-5.3.2-1/src/rcl/init_options.c:75
```

Do you know what could be causing this ?

### Additional context

I first had a qt issue that I solved by copying overlays for similar issue :

```diff
diff --git a/distros/distro-overlay.nix b/distros/distro-overlay.nix
index 3a5c29c15..68414fd9c 100644
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -312,6 +312,17 @@ let
         wrapQtApp "$out/lib/rviz/rviz"
       '';
     });
+    
+    rviz2 = rosSuper.rviz2.overrideAttrs ({
+      nativeBuildInputs ? [], postFixup ? "", ...
+    }: {
+      dontWrapQtApps = false;
+      nativeBuildInputs = nativeBuildInputs ++ [ self.qt5.wrapQtAppsHook ];
+      postFixup = postFixup + ''
+        wrapQtApp "$out/lib/rviz2/rviz2"
+      '';
+    });
+
 
     rviz-ogre-vendor = rosSuper.rviz-ogre-vendor.overrideAttrs ({
       preFixup ? "", ...
```

movefasta:
It should be fixed in https://github.com/lopsided98/nix-ros-overlay/pull/256

PR #256, opened 2023-03-28
Make rviz2 work

wentasah:
These changes allow me to run `rviz2`.

Issue #248 is related, though not completely solved by these changes.


wentasah:
Somehow, I no longer need to overlay `rviz-ogre-vendor` so I removed that from the commit and rebased the rest onto `master`.

wentasah:
It seems the last comment is not true. I tested it in an environment, which had LD_LIBRARY_PATH with needed libraries.

wentasah:
So I added `rviz-ogre-vendor` fix back. Not it's in a separate commit and all commits have commit messages explaining why it is needed. I also added an ability to run `rviz2` with `nix run`.

PR #257, opened 2023-03-31
flake: Configure the binary cache and add examples

wentasah:
This add the examples to the flake and automatically enable the binary cache. With this, one can simply run the example as:
```sh
nix develop github:wentasah/nix-ros-overlay/configure-cache-in-flake#example-turtlebot3-gazebo
```


movefasta:
@lopsided98 
Can we give some users permission to update the `staging` branch to update the overlay faster? Some active users (like @wentasah and some others you probably know from commit history) are already qualified enough to review and merge. I think that will be good for community development.
The `master` branch will be still updated by the maintainer and safe.

Issue #161, opened 2022-01-07
Gazebo crash on startup


erdnaxe:
I installed `nix-ros-overlay.rosPackages.noetic.gazebo`. When I run `gazebo --verbose`, I get:
```
Gazebo multi-robot simulator, version 11.9.0
Copyright (C) 2012 Open Source Robotics Foundation.
Released under the Apache 2 License.
http://gazebosim.org

[Msg] Waiting for master.
[Msg] Connected to gazebo master @ http://127.0.0.1:11345
[Msg] Publicized address: 192.168.56.1
Gazebo multi-robot simulator, version 11.9.0
Copyright (C) 2012 Open Source Robotics Foundation.
Released under the Apache 2 License.
http://gazebosim.org

[Msg] Waiting for master.
[Msg] Connected to gazebo master @ http://127.0.0.1:11345
[Msg] Publicized address: 192.168.56.1
[Msg] Loading world file [/nix/store/qgrfkh8bqddgzkawb8irnwrxxrkdq8q5-gazebo-11.9.0/share/gazebo-11/worlds/empty.world]
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
[ALSOFT] (EE) Failed to set real-time priority for thread: Operation not permitted (1)
[Err] [Scene.cc:227] Service call[/shadow_caster_material_name] timed out
[Err] [Scene.cc:249] Service call[/shadow_caster_render_back_faces] timed out
[Wrn] [Scene.cc:462] Ignition transport [/scene_info] service call failed, falling back to gazebo transport [scene_info] request.
[Wrn] [GuiIface.cc:120] Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.

[Err] [main.cc:37] Ogre Error:RuntimeAssertionException: Ogre/ShadowExtrudePointLight not found. Verify that you referenced the 'ShadowVolume' folder in your resources.cfg in initialise at /build/source/OgreMain/src/OgreShadowVolumeExtrudeProgram.cpp (line 70)
```

![gazebo_crash](https://user-images.githubusercontent.com/2663216/148610947-6f8d85eb-58ba-4ce9-a4e4-e891ca4b26eb.png)

[...]

PR #261, opened 2023-05-07
ogre1_9: init at 1.9.1

jtojnar:
Introduce version of Ogre removed from Nixpkgs.
Required by https://github.com/lopsided98/nix-ros-overlay/issues/161
Taken from https://github.com/NixOS/nixpkgs/commit/40f52d9003c5cb4929a346529d2ff1357bd89d6c

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -111,6 +111,8 @@ in {
     utils = self.ignition.utils1;
   };
 
+  ogre1_9 = self.callPackage ./ogre/1.9.x.nix { };
+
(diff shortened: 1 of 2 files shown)


kjeremy:
I think this is needed. See: https://github.com/lopsided98/nix-ros-overlay/commit/b1a844aed0b6614543199267303d5d7153785864#commitcomment-111528850

I'm upgrading our nixpkgs revision and just ran into this while attempting a switch to nixos 23.05 beta

PR #263, opened 2023-05-10
Add colcon extensions

wentasah:
This adds a few missing colcon extensions and the last commit enables already present extensions, which are enabled in typical colcon installatins (i.e. official ROS docker images).

I can imagine that the last commit, which enables parallel builds and hides build logs, might be controversial, because it could make debugging build failures harder. If this is the reason for those extension not being currently enabled, I would propose to introduce `colcon-full` with those extensions enabled. Then, we would recommend `colcon-full` for interactive sessions and keep `colcon` for Nix package building.


lopsided98:
I think it makes sense to match the `colcon-common-extensions` package as you have done.

Issue #75, opened 2021-01-21
ros2doctor not found


akru:
**Description**

Currently, some packages like `desktop` or [ros-core](https://github.com/lopsided98/nix-ros-overlay/blob/master/distros/foxy/ros-core/default.nix#L5) from ROS2 distros require a `ros2doctor` package that didn't present in the overlay.

 The reason I guess is this package included in the `ros2cli` set: https://index.ros.org/p/ros2doctor/

lopsided98:
`ros2doctor` depends on the `ifcfg` Python package, which is not available in nixpkgs. If you want `ros2doctor`, please submit a PR to nixpkgs to add `ifcfg`.

lopsided98:
https://github.com/NixOS/nixpkgs/pull/164515

wentasah:
Thanks @movefasta for mentioning this. I've sent a PR to update the ROS 2 example to use `ros-core`, which was previously uninstalable due to missing `ifcfg`. As far as I can see, `ros2doctor` is still missing in *foxy*. It's available in *humble* and *rolling* only.

lopsided98:
Foxy is EOL now so I think this can be closed.

PR #265, opened 2023-05-29
Update ros2-basic example to include ros-core and colcon

wentasah:
Previously, ros-core was not included presumably due to its dependency on ros2doctor, which depended on ifcfg package unavailable in nixpkgs. As pointed in https://github.com/lopsided98/nix-ros-overlay/issues/75#issuecomment-1508021227, this is no longer true. Since ros2cli version 0.17.0 (2022-01-25) ifcfg is no longer necessary as it was replaced by psutils [1]. This means that we can use ros-core without problems instead of manually specifying its dependencies such ros-environment, ros2topic, etc.

In addition to ros-core, colcon is a basic tool for ROS 2. Include it in the example too.

[1]: https://github.com/ros2/ros2cli/commit/cc5d6b1f9b26ada1a8f366ef7fe9534077c859c2

### DIFF
--- a/examples/ros2-basic.nix
+++ b/examples/ros2-basic.nix
@@ -8,9 +8,8 @@ mkShell {
   nativeBuildInputs = [
     (buildEnv {
       paths = [
-        ros-environment
-        ros2topic
-        ros2node
+        ros-core
+        colcon
         geometry-msgs
       ];
     })


hacker1024:
I was under the impression that Nix could pretty much remove the need to use colcon manually. What kinds of things is its direct use still needed for?

lopsided98:
I use it to build my own packages during development. But you're right, maybe it doesn't belong in this example.

wentasah:
I use it for development too. Not only for my own packages, but also for other packages/workspaces, which are not a part of ROS distributions. Writing Nix expressions for such projects is more work than simply running `colcon build`.

My reasoning for adding colcon to the example was that with it, it's easy to follow various ROS 2 tutorials.

PR #269, opened 2023-06-27
Make buildEnv more flexible

hacker1024:
This PR changes `buildEnv` to:

- Stop passing `wrapPrograms` to `buildEnv` from Nixpkgs, preventing errors from occuring
- Allow adding custom values to `postBuild` and `passthru`
- Allow supplying custom arguments to `makeWrapper` (e.g. to set the default RMW implementation to work around #45)


hacker1024:
@lopsided98 what are your thoughts on this change?

PR #270, opened 2023-06-30
Make binaries from ament_python packages available in $out/lib/$pkg

### LINKED ISSUE #230: ros2 run does not work with python packages.
I installed `rqt-robot-monitor` into a nix shell, but I am unable to run the executable.

- `ros2 pkg list | grep rqt_robot_monitor` shows that the package is expected.
- `ros2 run rqt_robot_monitor rqt_robot_monitor` reports `No executable found`
- I am able to run the executable using `python3 -m rqt_robot_monitor.robot_monitor`

I observer this behavior with multiple python packages.

Happy to help if you can point me in the right direction, thanks!

wentasah:
ROS 2 expects the binaries to be installed in $out/lib/$pkg (where $pkg is ROS package name) rather than in $out/bin, where buildPythonPackage installs them. Without binaries in $out/lib/$pkg, commands such as `ros2 run` and `ros2 launch` do not work.

The reason of this mismatch is difference in how colcon and buildPythonPackage build and install packages.

colcon builds Python packages by calling:

    setup.py egg_info ... build ... install ...

This command reads setup.cfg, which requests installation of binaries under lib, i.e., it contains:

    [install]
    install_scripts=$base/lib/my_package_name

On the other hand buildPythonPackage (used by ament_python packages in nix-ros-overlay) calls:

    setup.py bdist_wheel
    pip install dist/*.whl

This procedure ignores the content of setup.cfg.

This commit works around the differences by manually linking the binaries to the $out/lib/$pkg directory.

Fixes #230

### DIFF
--- a/distros/build-ros-package/default.nix
+++ b/distros/build-ros-package/default.nix
@@ -1,5 +1,6 @@
 { stdenv, lib, pythonPackages, rosDistro, rosVersion }:
 { buildType ? "catkin"
+, pname
   # Too difficult to fix all the problems with the tests in each package
 , doCheck ? false
 # nixpkgs requires that either dontWrapQtApps is set or wrapQtAppsHook is added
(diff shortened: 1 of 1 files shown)


hacker1024:
Two points:

1. I think it would be better to move the binaries instead of linking them, as they're not normally supposed to be in `bin/` at all. This can cause unexpected collisions in `buildEnv`.
2. The dependency on `pname` is quite brittle. We use `name` instead of `pname` for our packages, and we also prefix them all with our organization name - this won't work in either of those cases. We have bash at our disposal - is it not possible to parse the `package.xml` or even `setup.cfg` during build time?

wentasah:
If I move `bin/*` to `lib/`, it moves only the wrappers. Original binaries that start with `.` will stay in `bin/` and have to stay there, because the wrapper refers to them under `bin/`. So I guess the collisions are likely to happen too. We might move the binaries before wrapping and then wrap them explicitly in lib, but it might be trickier to figure out which wrappers need to be applied there. I'd definitely need to understand more about wrapping logic in nixpkgs.

As for `pname` - I agree. I also have a few packages that use `name` directly. I was hoping, I'd be able to use `fromTOML` to parse `setup.cfg`, but it fails due to missing quotes around string values. I'd like to avoid having a full parser in bash, but simple `sed` might be sufficient. Are you aware any package with complex `setup.cfg` where `sed` would fail?

hacker1024:
colcon uses a deprecated method of Python package installation, so this isn't an upstream Nixpkgs issue.

I have replicated the colcon build behaviour in #272, which I think is a more elegant solution to this problem.

PR #308, opened 2023-09-26
Fix hash for ogre on Iron

Programmerino:
It seems like the hash in the Iron `overrides.nix` file for ogre has become mismatched. This updates it to match the latest hash, after which it builds correctly.


lopsided98:
oh no... this means something in that mess is non-reproducible

hacker1024:
Perhaps we should merge first and ask questions later?

lopsided98:
It's not really inevitable; fixed output derivations work fine as long as you take care of all the non-reproducible parts.

In this case, I forgot that patching `Components/Overlay/CMakeLists.txt` would create a non-deterministic timestamp.

PR #335, opened 2023-12-04
buildEnv: Disable redundant fixup operations

hacker1024:
The fixupPhase is needed for shell hooks and input propagation, but other things like RPATH shrinking and shebang patching are not needed, as the original packages should have already been fixed up.

This significantly reduces `buildEnv` build times. The shebang patching in particular is quite slow.


wentasah:
This makes sense. I've verified that it doesn't break my (ROS 2) projects.

PR #338, opened 2023-12-19
Add packages to  pkgs to distro overlay

### LINKED ISSUE #150: legacyPackages lacks colcon and other tools
When using the flake for `nix-ros-overlay`, certain tools like `colcon` are not included under `legacyPackages`. I believe this is because you scope the export as `pkgs.rosPackages`.

koalp:
Hello, I share my workaround for #150 
I do not know what consequences it has but it  is what I am using to have access to `colcon`.
I have only tested with humble.

closes #150


wentasah:
Hi, I thinks that #481 allows doing the same in a cleaner way.

PR #341, opened 2024-01-02
buildEnv: Add ROS version environment variables to wrapper

hacker1024:
This PR adds some ROS version environment variables to the `buildEnv` wrapper. These variables are normally included by `nix-shell`, but are left out when building the wrapper.

ROS packages expect these variables to exist, and can fail when they are unset. `foxglove_bridge`, for example, crashes when `ROS_DISTRO` is unset.

See also:
- [REP 123: ROS_ETC_DIR, ROS_DISTRO environment variables and ROS_ROOT changes](https://ros.org/reps/rep-0123.html)
- [REP 149: Package Manifest Format Three Specification](https://ros.org/reps/rep-0149.html#environment-variables)

### DIFF
--- a/distros/build-env/default.nix
+++ b/distros/build-env/default.nix
@@ -10,7 +10,7 @@
 #
 # By default, all binaries in the environment are wrapped, setting the relevant
 # ROS environment variables, allowing use outside of nix-shell.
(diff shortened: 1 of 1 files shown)


wentasah:
This looks reasonable and it doesn't break my project.

PR #345, opened 2024-01-12
Add colcon-clean extension

nwn:
This adds the [colcon-clean](https://github.com/colcon/colcon-clean) extension that provides a nice command for workspace cleaning. The extension relies on the `scantree` Python package which isn't part of nixpkgs, so that package is specified manually as well. I'm not sure if that's how you'd typically do this or whether the dependency should be specified differently.

### DIFF
--- a/pkgs/colcon/clean.nix
+++ b/pkgs/colcon/clean.nix
@@ -1,0 +1,22 @@
+{ lib, buildPythonPackage, fetchPypi, colcon-core, scantree }:
+
+buildPythonPackage rec {
+  pname = "colcon-clean";
+  version = "0.2.0";
+
+  src = fetchPypi {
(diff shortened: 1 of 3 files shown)


wentasah:
I tested this in `nix repl` via:
```
:l .
:u with python3Packages; colcon-core.withExtensions [ colcon-argcomplete colcon-bash colcon-cmake colcon-core colcon-defaults colcon-devtools colcon-library-path  colcon-metadata colcon-notification colcon-output colcon-package-information colcon-package-selection colcon-parallel-executor colcon-python-setup-py colcon-recursive-crawl colcon-ros colcon-test-result colcon-zsh colcon-clean ]
colcon ...
```
Seems to work as expected.

Issue #343, opened 2024-01-08
Difficulty running turtlebot examples


grnnja:
Hi! I'm having difficulty running the example turtlebot flake command. It tries to find `opencv.cxxdev` but can't. Not sure if this is a problem on my end.

```bash
[prem@morokei:~]$ nix develop github:lopsided98/nix-ros-overlay#example-turtlebot3-gazebo
error:
       … while calling the 'derivationStrict' builtin

         at /derivation-internal.nix:9:12:

            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'nix-shell'
         whose name attribute is located at /nix/store/im9ymqc98dwazijqagl057yalg78lxx6-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'buildInputs' of derivation 'nix-shell'

         at /nix/store/im9ymqc98dwazijqagl057yalg78lxx6-source/pkgs/stdenv/generic/make-derivation.nix:395:7:

          394|       depsHostHost                = elemAt (elemAt dependencies 1) 0;
          395|       buildInputs                 = elemAt (elemAt dependencies 1) 1;
             |       ^
          396|       depsTargetTarget            = elemAt (elemAt dependencies 2) 0;

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'cxxdev' missing

       at /nix/store/rgrwzc3qyq0rlv6q7df3jdzaww4bzqgm-source/distros/noetic/cv-bridge/default.nix:19:42:

           18|   checkInputs = [ python3Packages.numpy rostest ];
           19|   propagatedBuildInputs = [ boost opencv opencv.cxxdev python3 python3Packages.opencv4 rosconsole sensor-msgs ];
             |                                          ^
           20|   nativeBuildInputs = [ catkin ];
```

with the nix shell example, it pulls the packages and compiles successfully but I am unable to run the demo.  I let it sit for a few minutes and nothing happened before I ctrl+c'd the process. Again I'm not sure if this is a problem on my end or not.

```bash
[prem@morokei:~]$ nix-shell \
  -I nix-ros-overlay=https://github.com/lopsided98/nix-ros-overlay/archive/master.tar.gz \
  --option extra-substituters 'https://ros.cachix.org' \
[...]

PR #351, opened 2024-02-10
fix cv-bridge: remove cxx output from opencv

drsk0:
Seems like the `opencv.cxxdev` output has been removed from the opencv package and is not necessary anymore for `cv-bridge`. With this change `nix develop github:lopsided98/nix-ros-overlay#example-turtlebot3-gazebo` works again. I also tested to build the `cv-bridge` package for the other distros.


lopsided98:
It's the other way around; the cxxdev was added in a newer nixpkgs version than was pinned, but this is now fixed. See #343.

drsk0:
I see, thanks for fixing it anyway! By the way, the turtlebot example also hangs for me. Thanks for the effort you but into this, I'm happy I have found nix derivations for ROS.

PR #366, opened 2024-03-04
- moveit-kinematics: add moveit-ros-planning to build inputs

akssri-sony:
Build errors out otherwise.


lopsided98:
That file is auto-generated and any changes will be overwritten. These kind of changes should go in the distro overlay. I went ahead and did it here: https://github.com/lopsided98/nix-ros-overlay/commit/fc2608672ea7acf465e3c92b31142573c2f22cfe

akssri-sony:
Ah I see; thanks!

Issue #319, opened 2023-10-15
ros-gz Fails to build, ros-gz-bridge ros-gz-images, etc.. don't exist


tmayoff:
```
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'ros-devenv'
         whose name attribute is located at /nix/store/rkkiiyv317i5c9iwqz8cz7prvd04k08k-source/pkgs/stdenv/generic/make-derivation.nix:300:7

       … while evaluating attribute 'buildInputs' of derivation 'ros-devenv'

         at /nix/store/rkkiiyv317i5c9iwqz8cz7prvd04k08k-source/pkgs/stdenv/generic/make-derivation.nix:347:7:

          346|       depsHostHost                = lib.elemAt (lib.elemAt dependencies 1) 0;
          347|       buildInputs                 = lib.elemAt (lib.elemAt dependencies 1) 1;
             |       ^
          348|       depsTargetTarget            = lib.elemAt (lib.elemAt dependencies 2) 0;

       error: evaluation aborted with the following error message: 'Function called without required argument "ros-gz-bridge" at /nix/store/dc7sviyvcgif1ss22kgx5pvvx28dks32-source/distros/humble/ros-gz/default.nix:5, did you mean "ros-ign-bridge"?'
```

lopsided98:
New Gazebo (as opposed to classic) and all its Ignition dependencies need to be packaged manually. This is quite a bit of work, and I started it at one point but haven't had time to finish it.

muellerbernd:
I've tried to get ignition gazebo running. You can find my sources [here](https://github.com/muellerbernd/gazebo-sim-overlay). In that repo I tried to mimic the structure we are using in the nix-ros-overlay. My repo is still WIP but I managed to compile [Gazebo Sim Harmonic](https://github.com/gazebosim/gz-sim/releases/tag/gz-sim8_8.0.0) and all it's dependencies. The GUI is not working right now:
```bash
gz sim 
Library error: libgz-tools2-backward.so not found. Improved backtrace generation will be disabled
[...]

PR #370, opened 2024-03-06
gazebo: add wrapGAppsHook

MatthewCroughan:
makes the derivation more pure, without this it leads to (gzclient:1568911): GLib-GIO-ERROR **: 15:37:29.803: No GSettings schemas are installed on the system, unless you happen to be lucky enough to have the right files in your environment


muellerbernd:
@MatthewCroughan I have created a repo for the gazebo simulators [here](https://github.com/muellerbernd/gazebo-sim-overlay). The old gazebo (the one used for ros1) works fine and I have adapted your changes proposed in this PR. Currently I am trying to get the new gazebo (the one used in ros2) running. #319 documents some things we tried. Maybe you are also interested in getting the new gazebo running and maybe you can help us there.

PR #374, opened 2024-03-21
- humble/ompl: patch the .pc file with cmake full paths

akssri-sony:


wentasah:
Hi, thanks for the patch. This this has already been [fixed upstream](https://github.com/ompl/ompl/pull/1101), but the fix didn't propagate to ROS releases yet. It would be better to apply the patch to all ROS 2 releases in `ros2-overlay.nix` instead of just to humble. I tested that ompl in humble, iron and rolling is broken.

akssri-sony:
Ah, I was going to make a upstream PR; but good to know that it's fixed. 

I've moved the patch to ros2-overlay.nix. I assume this'd break rolling and so I've removed it in ```rolling/overrides.nix```.
(Can you please check the rolling build ? A bit late on my end now.)

distros/ros2-overlay.nix
```diff
@@ -77,6 +77,12 @@ rosSelf: rosSuper: with rosSelf.lib; {
     ];
   });
 
+  ompl = rosSuper.ompl.overrideAttrs ({
+    ...
+  }: {
+    patches = [ ./ompl_ros2.patch ];
```
wentasah, line 83: One more thing (sorry for not noticing this earlier): You should append to existing patches (if they are added in the future).
```suggestion
    patches ? [], ...
  }: {
    patches = patches ++ [ ./ompl_ros2.patch ];
```

akssri-sony: fixed.

lopsided98:
Noetic needs this as well, and I also I prefer to fetch patches from their upstream source where possible. I went ahead and did that in dc5a4a3fa52f98972708abb996e73e843c08a09b

akssri-sony:
ah, makes sense; thanks.

PR #377, opened 2024-03-26
Adding easy_handeye2 packages

Thieso:
Hello, 

I wanted to use the easy_handeye2 package https://github.com/marcoesposito1988/easy_handeye2 and it uses a rqt plugin for its GUI. This makes it impossible to just put it in the workspace and build it in the `nix develop` shell because then it cannot find the QT plugins. Thats why I added them to the distro here. 

However, the original repo https://github.com/marcoesposito1988/easy_handeye2 builds 2 packages, one for the rqt plugin and the other for the messages. Is it possible to change the build folder accordingly? Because otherwise the `setup.py` is not found because it is not in the parent directory. Right now I just split it up into two directories. 

Furthermore, for the rqt plugins one has to copy the xml files `plugin_calibrator.xml` and `plugin_evaluator.xml` into the local ROS install folder but I have no idea how to do this in nix. Do you have experience with this? For now I copied it by hand. 

So when I can improve these two things everything works great.

### DIFF
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -239,6 +239,16 @@ let
       '';
     });
 
+    easy-handeye2 = rosSuper.easy-handeye2.overrideAttrs ({
+      nativeBuildInputs ? [], postFixup ? "", ...
+    }: {
+      dontWrapQtApps = false;
+      nativeBuildInputs = nativeBuildInputs ++ [ self.qt5.wrapQtAppsHook ];
+      postFixup = postFixup + ''
+        wrapQtApp "$out/lib/easy_handeye2/rqt_calibrator.py"
+      '';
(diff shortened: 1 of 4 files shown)


lopsided98:
Sorry, this overlay only supports packages that are published as part of rosdistro, which enables the package definitions be auto-generated with superflore. If you want to have it included, you need to get the maintainers to publish the package.

You can package it manually yourself outside of the repo though, pretty much as you have already done. You can build from a subdirectory using `sourceRoot`, probably something like `sourceRoot = "source/easy_handeye2_msgs";`. The XML files are included in `data_files` in `setup.py`, so I would expect that to just work.

Thieso:
Ok that makes sense. Thanks for your help

PR #394, opened 2024-04-16
Fix sophus build errors for humble and iron

### LINKED ISSUE #393: Sophus doesn't build
I'm trying to add the Sophus library to my package, which doesn't seem to build for certain ROS distros. In my case, it is specifically the noetic version that doesn't build, but several others seem not to work either. 

I've done some rudimentary testing, and have these results: 

foxy: fails
humble: fails
iron: fails
noetic: fails
rolling: builds

It seems that the Sophus package builds the tests and examples, and this is where it crashes. Is there any way of disabling these?

wentasah:
We apply the patch from the rolling version, which builds fine already now. The patch does not apply to foxy and noetic versions so we don't patch them. Foxy is EOL. If someone needs noetic version, more effort is needed.

Fixes #393

### DIFF
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -134,6 +134,18 @@ in with lib; {
     sha256 = "1iv6k0dwdzg5nnzw2mcgcl663q4f7p2kj7nhs8afnsikrzxxgsi4";
   };
 
+  sophus = rosSuper.sophus.overrideAttrs ({
+    patches ? [], ...
+  }: {
+    patches = patches ++ [
+      # Remove -Werror from the compile flags.
+      (self.fetchpatch {
+        url = "https://github.com/clalancette/sophus/commit/13339e883e627dc62e5a3e8d13919173051ea139.patch";
+        hash = "sha256-vY99tXRcrWxcztva+wQiOw50PwLePrvaSVZ/lerX1Kg=";
+      })
+    ];
+  });
+
   urdfdom = rosSuper.urdfdom.overrideAttrs ({
     patches ? [], ...
   }: {
--- a/distros/iron/overrides.nix
+++ b/distros/iron/overrides.nix
@@ -118,6 +118,18 @@ in with lib; {
     '';
   });
 
+  sophus = rosSuper.sophus.overrideAttrs ({
+    patches ? [], ...
+  }: {
+    patches = patches ++ [
+      # Remove -Werror from the compile flags.
+      (self.fetchpatch {
+        url = "https://github.com/clalancette/sophus/commit/13339e883e627dc62e5a3e8d13919173051ea139.patch";
+        hash = "sha256-vY99tXRcrWxcztva+wQiOw50PwLePrvaSVZ/lerX1Kg=";
+      })
+    ];
+  });
+
   urdfdom = rosSuper.urdfdom.overrideAttrs ({
     patches ? [], ...
   }: {
--- a/distros/noetic/overrides.nix
+++ b/distros/noetic/overrides.nix
@@ -83,4 +83,12 @@ rosSelf: rosSuper: with rosSelf.lib; {
     url = "https://github.com/fmtlib/fmt/releases/download/9.1.0/fmt-9.1.0.zip";
     sha256 = "sha256-zOtMuTZuGKV0ISjLNSTOX1Doi0dvHlRzekf/3030yZY=";
   };
+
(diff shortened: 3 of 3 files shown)


martiege:
Nice! Do you have any idea why the noetic version doesn't work?

wentasah:
Do you need noetic? I didn't look into the details. I assume the reason will be the same (i.e. newer compiler in nixpkgs generates more warnings). The solution would be also the same - remove `-Werror` from `CMakeLists.txt`, but that file looks differently in noetic and the patch doesn't apply. A different patch needs to be made for noetic.

martiege:
Here is a quick patch doing that for noetic (on the upstream here: https://github.com/stonier/sophus/tree/release/1.1.x)

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 26f0b3d..48fa083 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,11 +16,11 @@ set(CMAKE_CXX_STANDARD 11)
 IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
    SET(CMAKE_CXX_FLAGS_DEBUG  "-O0 -g")
    SET(CMAKE_CXX_FLAGS_RELEASE "-O3")
-   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wextra  -Wno-deprecated-register -Qunused-arguments -fcolor-diagnostics")
+   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra  -Wno-deprecated-register -Qunused-arguments -fcolor-diagnostics")
 ELSEIF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
    SET(CMAKE_CXX_FLAGS_DEBUG  "-O0 -g")
    SET(CMAKE_CXX_FLAGS_RELEASE "-O3")
-   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wextra -std=c++11 -Wno-deprecated-declarations -ftemplate-backtrace-limit=0")
+   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -std=c++11 -Wno-deprecated-declarations -ftemplate-backtrace-limit=0")
    SET(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_DEBUG} --coverage -fno-inline -fno-inline-small-functions -fno-default-inline")
    SET(CMAKE_EXE_LINKER_FLAGS_COVERAGE "${CMAKE_EXE_LINKER_FLAGS_DEBUG} --coverage")
    SET(CMAKE_SHARED_LINKER_FLAGS_COVERAGE "${CMAKE_SHARED_LINKER_FLAGS_DEBUG} --coverage")

```

martiege:
But if this is all that is needed to do, and the error seems to stem from building the tests and examples, wouldn't it be better to turn off those options?

martiege:
Btw, while it is annoying that some of the ros-versions of sophus don't work, the library itself can be built and used with no issue. I just copied the derivation from nixpkgs unstable: 
```nix
{ lib, stdenv, eigen, fmt, fetchFromGitHub, cmake }:
stdenv.mkDerivation rec {
  pname = "sophus";
  version = "1.22.10";

  src = fetchFromGitHub {
    owner = "strasdat";
    repo = "Sophus";
    rev = "${version}";
    hash = "sha256-TNuUoL9r1s+kGE4tCOGFGTDv1sLaHJDUKa6c9x41Z7w=";
  };

  nativeBuildInputs = [ cmake ];

  propagatedBuildInputs = [ eigen fmt ];

  cmakeFlags = [
    (lib.cmakeBool "BUILD_SOPHUS_TESTS" false)
    (lib.cmakeBool "BUILD_SOPHUS_EXAMPLES" false)
  ];
}
```

Then add it to the code I want like a normal library in cmake: 

```
find_package(Sophus REQUIRED)
target_link_libraries(target Sophus::Sophus)
```

wentasah:
> But if this is all that is needed to do, and the error seems to stem from building the tests and examples, wouldn't it be better to turn off those options?

I'd prefer to build tests and examples, because it may allow discovering potential problems in the future. We just need to remove `-Werror`, because we use a newer compiler than upstream.

I've updated the commit to also patch Noetic. Now, all sophus versions build fine.

wentasah:
This is no longer needed. Closing.

PR #397, opened 2024-04-23
Add fixupPhase flag to fix the cmake copy issue

### LINKED ISSUE #391: Moveit Packages in Iron: Ros-env cmake copy issue
Hello, 

I am trying to install MoveIt in Ros2 Iron but for some package I get the error: 

```
Running phase: fixupPhase
> moving /nix/store/rlzhp0g8haspl9w4a3dcp54zbhxpwgvq-ros-env/lib64/* to /nix/store/rlzhp0g8haspl9w4a3dcp54zbhxpwgvq-ros-env/lib
> mv: not replacing '/nix/store/rlzhp0g8haspl9w4a3dcp54zbhxpwgvq-ros-env/lib/cmake'
```

The packages that encounter this issue are:

- `moveit-ros`
- `moveit-servo`
- `moveit-planners`

and possibly more which I have not tested. 

Do you know what causes this error and how I may help to fix it? 

Thank you

Thieso:
Fixes #391 

But does this break anythink else? Is this a good fix?

### DIFF
--- a/distros/build-env/default.nix
+++ b/distros/build-env/default.nix
@@ -108,5 +108,6 @@ let
     noAuditTmpdir = true;
     dontGzipMan = true;
     dontPatchShebangs = true;
+    dontMoveLib64 = true;
   });
 in env


wentasah:
I think this should not break anything. Packages from this overlay are built with `moveLib64` hook enabled so `lib64` should always be a symlink to `lib`. `buildEnv` just creates one more level of symlinks, which should be fine.

The only downside I can see is that this approach may consume more i-nodes for symlinks than necessary. Creating one symlink from `...ros-env/lib64` to `lib` should be sufficient, but this creates full tree of new symlinks. But I don't think it's a problem for anybody.

hacker1024:
I'm a little concerned about potential performance impacts here. `buildEnv` is already quite slow, especially on devices with low-end storage.

Thieso:
Fair, so would there be an easier way to achieve this? Maybe force overwriting of redundant files?

Thieso:
Would be great if this could be merged :) Thanks

PR #409, opened 2024-05-23
Allow gazebo to build against graphviz 10

kjeremy:
Unstable ships with graphviz 10 instead of 9. Allow gazebo to compile against the newer version as well.


pkgs/gazebo/default.nix
```diff
@@ -18,6 +18,14 @@ mkDerivation rec {
     hash = "sha256-fphCwEbJ4HVTVbJ0wkCoq79Olivnznt/WRlOX0tYT0U=";
   };
 
+  patches = [
+    # Allow building with graphviz 10+
+    (fetchpatch {
+      url = "https://github.com/gazebosim/gazebo-classic/pull/3373.patch";
```
wentasah, line 24:
```suggestion
      url = "https://github.com/gazebosim/gazebo-classic/commit/660ae15f4af1f5ea8d9d50d4a24e4d91a94f9c2d.patch";
```

I think @lopsided98 prefers linking to patches via commit hashes rather than via PR numbers. The reason is probably that the PR diff can be easily changed, leading to hash mismatches.

PR #410, opened 2024-05-24
Add ROS 2 Jazzy Jalisco

wentasah:
The content of `jazzy/overrides.nix` is just a copy of `rolling/overrides.nix`.

At least `ros-core` builds successfully (tested on files generated by superflore today; not yet in this repo).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,6 +11,7 @@ jobs:
           - foxy
           - humble
           - iron
+          - jazzy
           - rolling
         system:
           - x86_64-linux
--- a/distros/default.nix
+++ b/distros/default.nix
@@ -29,6 +29,12 @@ self: super: {
       python = self.python3;
     } self super;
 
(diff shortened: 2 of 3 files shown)


hacker1024:
Perhaps we should use Python 3.12 here as well? [REP 2000](https://www.ros.org/reps/rep-2000.html#jazzy-jalisco-may-2024-may-2029) uses a range of 3.7 - 3.12, depending on the platform.

wentasah:
I'm not sure. I think that historically, the `python` attribute was used in this repo only to distinguish between Python 2 and 3. Given that Jazzy will be supported 5 years, it will likely be longer than how long will Python 3.12 stay in nixpkgs. Therefore, keeping this overlay up-to-date with what's the default in nixpkgs seems more practical to me.

hacker1024:
Fair enough. 3.12 will also become the default at some point in the near future: https://github.com/NixOS/nixpkgs/pull/320924

Issue #404, opened 2024-05-02
workaround included: rqt won't load cpp plugins due to SIP not loading even though Shiboken/PySide should be used


Pleune:
It seem like qt_qui_cpp is keeping rqt_image_view from loading into the rqt plugins list. It looks like after d3c90ec, Pyside/Shiboken should be used instead of PyQt/SIP,  but python_qt_binding is still preferring SIP.  Interestingly, when running plugins directly, they seem to work. For example `ros2 run rqt_image_view rqt_image_view`, instead of trying to add it to `rqt`.

**Error when running `rqt`**
```
Could not import "pyqt" bindings of qt_gui_cpp library - so C++ plugins will not be available:
Traceback (most recent call last):
  File "/nix/store/hyf3pk2c0cglgfzb064nlxg76ldn6zwx-ros-env/lib/python3.11/site-packages/qt_gui_cpp/cpp_binding_helper.py", line 43, in <module>
    from . import libqt_gui_cpp_sip
ImportError: cannot import name 'libqt_gui_cpp_sip' from 'qt_gui_cpp' (/nix/store/hyf3pk2c0cglgfzb064nlxg76ldn6zwx-ros-env/lib/python3.11/site-packages/qt_gui_cpp/__init__.py)
```

Which shows that QT_BINDING is being chosen as pyqt. But as libqt_gui_cpp_sip is not in the qt_qui_cpp folder, due to the linked commit. Reverting that commit does not help, because SIP still is broken and does not compile. I have not investigated trying to fix that compile issue.

**/nix/store/hyf3pk2c0cglgfzb064nlxg76ldn6zwx-ros-env/lib/python3.11/site-packages/qt_gui_cpp/cpp_binding_helper.py:**
```
  33   │ from python_qt_binding import QT_BINDING
  34   │ from python_qt_binding.QtCore import qWarning
  35   │
  36   │
  37   │ try:
  38   │     if QT_BINDING == 'pyside':
  39   │         from . import libqt_gui_cpp_shiboken
  40   │         qt_gui_cpp = libqt_gui_cpp_shiboken.qt_gui_cpp
  41   │
  42   │     elif QT_BINDING == 'pyqt':
  43   │         from . import libqt_gui_cpp_sip
  44   │         qt_gui_cpp = libqt_gui_cpp_sip.qt_gui_cpp
  45   │
  46   │     else:
  47   │         raise ImportError('Qt binding name "%s" is unknown.' % QT_BINDING)
```

[...]

PR #414, opened 2024-05-30
ros2 python-qt-binding sip4 patch to fix apps like rqt

Pleune:
Without this patch, on WSL, and I believe all non-darwin systems, rqt will be unable to load the image_viewer plugin. See #404 and https://github.com/lopsided98/nix-ros-overlay/issues/404#issuecomment-2092797850

This could be selectively applied to rqt, similar to the ros1 overlay rviz patch. But I think it would be best to patch for all dependencies. This should not break anything, as it does not affect whether or not pyside or pyqt is chosen as the backend. It will simply allow the pyqt backend to function.

An alternative approach could be taken (and also works) where python-qt-bindings is patched to always choose pyside, but I actually see better framerate with the image_viewer plugin with pyqt (at least on WSL2).

I have tested this with rqt in ros2/humble.

### DIFF
--- a/distros/ros2-overlay.nix
+++ b/distros/ros2-overlay.nix
@@ -112,8 +112,7 @@ rosSelf: rosSuper: with rosSelf.lib; {
     postPatch ? "", ...
   }: {
     propagatedNativeBuildInputs = with rosSelf.pythonPackages;
-      (rosSelf.lib.subtractLists [ shiboken2 pyside2 ] propagatedNativeBuildInputs)
-      ++ [ sip_4 ];
+      propagatedNativeBuildInputs++ [ sip_4 ];
     postPatch = ''
       sed -e "1 i\\import PyQt5" \
           -e "s#sipconfig\._pkg_config\['default_mod_dir'\], 'PyQt5'#PyQt5.__path__[0]#" \
--- a/distros/ros2-overlay.nix
+++ b/distros/ros2-overlay.nix
@@ -113,11 +113,11 @@ rosSelf: rosSuper: with rosSelf.lib; {
   }: {
     propagatedNativeBuildInputs = with rosSelf.pythonPackages;
       propagatedNativeBuildInputs++ [ sip_4 ];
-    postPatch = ''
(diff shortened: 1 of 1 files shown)


Pleune:
This breaks the compilation on rolling and Jazzy, because the sip4 patch is applied internally in the python-qt-binding package at configure time. See https://github.com/ros-visualization/python_qt_binding/blob/1fbfb441b86dfc9df252fb7e4a7c0d736a5d1cbe/cmake/sip_configure.py#L143.

The added propagatedbuilddependency on sip_4 causes dependent packages to call the cmake helper file which un turns will try to run sup_configure.py and prevent a successful build. 

```
[ 42%] Built target ament_cmake_python_build_qt_gui_cpp_egg
[ 45%] Linking CXX shared library libqt_gui_cpp.so
[ 45%] Built target qt_gui_cpp
[ 48%] Running Shiboken generator for libqt_gui_cpp Python bindings...
[ 51%] Running SIP generator for qt_gui_cpp_sip Python bindings...
Traceback (most recent call last):
  File "/nix/store/yf6426f0j4p0l2q3lfq5swazng68ax5p-ros-rolling-python-qt-binding-2.3.0-r1/share/python_qt_binding/cmake/sip_configure.py", line 143, in <module>
    with open(os.path.join(tmpdirname, 'QtCore', 'QtCoremod.sip'), 'w') as outfp:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/build/tmp47o8fy9y/QtCore/QtCoremod.sip'
make[2]: *** [src/qt_gui_cpp_sip/CMakeFiles/libqt_gui_cpp_sip.dir/build.make:103: sip/qt_gui_cpp_sip/Makefile] Error 1
make[1]: *** [CMakeFiles/Makefile2:279: src/qt_gui_cpp_sip/CMakeFiles/libqt_gui_cpp_sip.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

PR #415, opened 2024-05-30
Ros1 python-qt-binding make pyqt patch global

Pleune:
For consistency with #414. I have not tested this, as right now I cannot build my ros1 flake at the moment due to failing to download `https://github.com/ros-gbp/bond_core-release/archive/release/noetic/bondcpp/1.8.6-1.tar.gz`... which I don't think is related but I don't have time to figure it out at the moment.


Pleune:
#404

PR #416, opened 2024-06-06
Python qt bindings support both

Pleune:
I have been bounding around trying to get all of the pyqt stuff to work across all of the distros for a while now... and I think this is my final solution. It seems to keep everything happy, and works even for the most current versions of everything, without requiring any huge overrides, or any patches that will break when files are updated upstream.

Some packages need specifically pyqt or pyside to be loaded by python-qt-binding. This changes the overrides to support both.

Annoyingly, in recent python-qt-binding releases, sip_configure.py, the cmake helper script, rewrites QtCoremod.sip, which is outside of the nix build area and causes a build failure. See https://github.com/ros-visualization/python_qt_binding/commit/e78372fd63eda527c9fad5fcdab8ca31eb3f36d2

This is the same patch that is applied from the AUR here. So sed patch out these changes. Then move the ros1 rviz patch to distro-overlay.nix to apply to everything.

Tested with:
 - ROS2 rolling using rqt_image_viewer
 - ROS2 humble using rqt_image_viewer
 - ROS1 noetic using rqt_image_viewer and rziv
 
 See #404

### DIFF
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -178,12 +178,23 @@ let
 
     pr2-tilt-laser-interface = patchBoostSignals rosSuper.pr2-tilt-laser-interface;
 
-    python-qt-binding = rosSuper.python-qt-binding.overrideAttrs ({
-      propagatedNativeBuildInputs ? [], ...
+    python-qt-binding = (rosSuper.python-qt-binding.override {
+      python3Packages = rosSelf.python3Packages.overrideScope (pyFinal: pyPrev: {
+        pyqt5 = pyPrev.pyqt5.overrideAttrs ({
+          patches ? [], ...
+        }: {
+          patches = patches ++ [ (self.fetchpatch {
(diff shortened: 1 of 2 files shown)


Pleune:
this patch is still needed for some QUI apps to work, including rqt, in the newly generated jazzy and rolling distros.

Pleune:
shiboken2 is now marked broken with python312 https://github.com/NixOS/nixpkgs/commit/603c0346d1dcdc329c4cf423a5cce510f1db3b57

Pleune:
ROS2 meta ticket involving upgrade from qt5 -> qt6, which would solve these 3.12 compatibility issues https://github.com/ros2/ros2/issues/1452

Pleune:
Submitted upstream changes to sip4 and pyside2 to allow it to work with Python312: https://github.com/NixOS/nixpkgs/pull/327976

With this I have rqt still working with both the SIP or PySide backends with newest nixpkgs/master.

Pleune:
Submitted PR upstream to update sip4 and pyside2/shiboken2 to support Python312. https://github.com/NixOS/nixpkgs/pull/327976

With these changes, I can now open and use rqt with all the plugins with 3.12 from current nixpkgs/master, using either backend.

kjeremy:
@lopsided98 is https://github.com/NixOS/nixpkgs/pull/327976 in your fork?

PR #417, opened 2024-06-07
ament-cmake-vendor-package: don't 'fix' cmake files

Pleune:
the ament cmake vendoring works by downloading files into the `ros2_ws/install/opt/<package>` folder, which is a *local* path. It then sets the `CMAKE_PREFIX_PATH` to point into this directory. However, the normal nixpkgs cmake setup renames the /opt/ chunk with /var/empty, which then breaks the CMAKE_PREFIX_PATH setup. since this is a local install, just dont "fix" the cmake paths.

I noticed this by trying to build https://github.com/ros2/rmw_zenoh, which uses the ament cmake vendoring, and therefore just simply does not build without this patch.

I had to apply this in the distro overlay files, as the vendoring package does not exist in foxy.

### DIFF
--- a/distros/ros2-overlay.nix
+++ b/distros/ros2-overlay.nix
@@ -12,6 +12,13 @@ rosSelf: rosSuper: with rosSelf.lib; {
     outputs = [ "out" "dev" ];
   });
 
+  ament-cmake-vendor-package = rosSuper.ament-cmake-vendor-package.overrideAttrs ({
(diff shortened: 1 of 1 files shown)


Pleune:
Rebased and add patch for Jazzy. This is still needed to support rmw_zenoh.

lopsided98:
This should go in `ros2-overlay.nix`

Pleune:
Updated to apply in `ros2-overlay.nix`

PR #418, opened 2024-06-17
Update all VRPN packages to a working version

Pleune:
Importantly, includes the changes under commits titled "deprecate CMAKE_HAVE_THREADS_LIBRARY", which allow its FindVRPN.cmake file to work on the newest cmake versions.

Other changes that are unique to the ros2-gbp/vrpn-release repo are needed, so don't pull from upstream.

### DIFF
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -387,6 +387,14 @@ let
     };
 
     urdf = patchBoostPython rosSuper.urdf;
+
+    vrpn = rosSuper.vrpn.overrideAttrs { 
+      src = self.fetchurl {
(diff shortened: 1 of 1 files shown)


wentasah:
Hi! Can you be more specific what is not working in this package? AFAICS, the package compiles fine. So I guess the problems are at run time. Does to problem appear in all ROS versions or only selected ones?

Pleune:
The vrpn package does actually compile fine, but the FindVRPN.cmake file it installs does not work with very recent versions of cmake, causing dependent packages from building.

This is because the mentioned used variable was undocumented, and was also removed in a recent cmake version. There is simply another equivalent and properly supported variable that should be used instead. The last few versions of the vrpn ros releases have a patch applied for this, even though it is not actually used until jade it seems. My guess is that Ubuntu/Debian do not use new enough cmake yet for it to matter. I did not bother to check what version of cmake is breaking.

lopsided98:
Jazzy and rolling have this fix now, so I just applied the relevant patch for humble.

Issue #419, opened 2024-06-20
Advice on adding macOS support to the nix-ros-overlay


hegde-atri:
I have an access to macOS(arm64) laptop in which I'd like to use ROS2 Humble.

I want to contribute to this project, but being new to nix & nix flakes I would like pointers to go down or to work on.

If there are any contribution guidelines I missed do let me know :)

hegde-atri:
Running the example `nix develop lopsided....` yields this output currently

```sh
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:9:12:
            8|
            9|   strict = derivationStrict drvAttrs;
             |            ^
           10|

       … while evaluating derivation 'nix-shell'
         whose name attribute is located at /nix/store/snp5f95xhkz89kdw9703kana3m6xfd0a-source/pkgs/stdenv/generic/make-derivation.nix:352:7

       … while evaluating attribute 'buildInputs' of derivation 'nix-shell'
         at /nix/store/snp5f95xhkz89kdw9703kana3m6xfd0a-source/pkgs/stdenv/generic/make-derivation.nix:399:7:
          398|       depsHostHost                = elemAt (elemAt dependencies 1) 0;
          399|       buildInputs                 = elemAt (elemAt dependencies 1) 1;
             |       ^
          400|       depsTargetTarget            = elemAt (elemAt dependencies 2) 0;

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: Package ‘ogre-1.9.1’ in /nix/store/zv36pjyxlsg6il3k6zhc1wdmgclwqqw9-source/pkgs/ogre/1.9.nix:86 is not available on the requested hostPlatform:
         hostPlatform.config = "aarch64-apple-darwin"
         package.meta.platforms = [
           "aarch64-linux"
           "armv5tel-linux"
           "armv6l-linux"
           "armv7a-linux"
           "armv7l-linux"
           "i686-linux"
           "loongarch64-linux"
           "m68k-linux"
           "microblaze-linux"
           "microblazeel-linux"
           "mips-linux"
           "mips64-linux"
           "mips64el-linux"
           "mipsel-linux"
           "powerpc64-linux"
           "powerpc64le-linux"
           "riscv32-linux"
           "riscv64-linux"
           "s390-linux"
           "s390x-linux"
           "x86_64-linux"
         ]
         package.meta.badPlatforms = [ ]
       , refusing to evaluate.

[...]

PR #421, opened 2024-07-06
Allow building ros-core for humble on MacOS

wentasah:
This fixes build issues on MacOS identified in #419. With these commits, it's possible to build `humble.ros-core` on MacOS. Unfortunately, building `ros-core` still fails on iron and rolling due to `lttng` being a dependency of `fastrtps`. I think this can be fixed, but will require more work, maybe in other repositories, so I don't try to fix it here.


hegde-atri:
this works completely fine for me (MacOS Sonoma 14.5) 👍🏼

PR #442, opened 2024-08-02
update foonathan-memory to fix memory allocation error on macOS

stevalkr:
fixes [foonathan::memory] Allocator foonathan::memory::memory_pool error, mentioned in https://github.com/lopsided98/nix-ros-overlay/issues/419#issuecomment-2196472985, see https://github.com/RoboStack/ros-humble/issues/32#issuecomment-2227591674

Tested on macOS 15.0-beta, needs to be verified on other versions and platforms

### DIFF
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -18,12 +18,14 @@ in with lib; {
   });
 
   # This is a newer version than the build system tries to download, but this
-  # version doesn't try to run host platform binaries on the build platform.
+  # version doesn't try to run host platform binaries on the build platform
+  # and fixes "Allocator foonathan::memory::memory_pool received invalid size"
+  # error on MacOS
   foonathan-memory-vendor = patchExternalProjectGit rosSuper.foonathan-memory-vendor {
     url = "https://github.com/foonathan/memory.git";
     originalRev = "v0.7-1";
-    rev = "v0.7-2";
-    fetchgitArgs.hash = "sha256-5nJNW0xwjSCc0Egq1zv0tIsGvAh1Xbnu8190A1ZP+VA=";
+    rev = "v0.7-3";
+    fetchgitArgs.hash = "sha256-nLBnxPbPKiLCFF2TJgD/eJKJJfzktVBW3SRW2m3WK/s=";
   };
 
   gazebo = self.gazebo_11;


hegde-atri:
Seems to fix the issue and works perfectly fine on MacOS Sonama 14.5 👍🏼

wentasah:
Thanks @etherswangel for finding the fix. The file `distros/humble/foonathan-memory-vendor/default.nix` is automatically generated so don't try to modify it. It would be overwritten on the next superflore run. The change in the `overrides.nix` file is sufficient. However, I would extend the comment above it to make it clear why we are using newer version than what's release in nix. I suggest changing the comment to:
```
  # This is a newer version than the build system tries to download, but this
  # version doesn't try to run host platform binaries on the build platform
  # and fixes "Allocator foonathan::memory::memory_pool received invalid size"
  # error on MacOS.
```

etherswangel:
Thanks, I've done what you said, please check.

wentasah:
Looks good. Could you also remove the first two commits? One way of doing that is using interactive rebase (e.g. `git rebase -i HEAD~3`) followed by `git push -f`.

etherswangel:
Could you use `Squash and merge`? This way I won't have to force push my fork and there'll be only one new commit in this repo. If it's not allowed in this repo I'll squash and force push

wentasah:
I cannot merge here, but hopefully @lopsided98 can.

In general I don't see a problem with forcepushing in your forks. Github can show the diff between the old and new version so it's easy to verify that nothing has changed. On the other hand, squashing requires extra effort from the maintainer to edit the commit message. In this case it's mostly irrelevant, but with longer commit messages it would be nice to save the maintainer the extra work.

etherswangel:
Sorry for the delayed response, I’ve been swamped this week. I’ve pushed the changes, but I think most developers may want the `—-force` not to be with them. The commit message is just there, you don’t need to edit anything. Just two clicks and you’re all set

wentasah:
@stevalkr I guess the above merge was not intentional. I'm not sure how Github's "squash & merge" behaves in this case, but yet another force-push to revert the merge might help :-).

stevalkr:
@wentasah Sorry that was a mistake. I forgot this is a branch in PR and still haven’t got merged. Well it's been half a month since my last force-push so I doubt it's getting merged... 

@lopsided98 would you take a look?

PR #445, opened 2024-08-03
support rviz2 for humble on macOS

stevalkr:
fixes rviz2 issue in https://github.com/lopsided98/nix-ros-overlay/issues/419#issuecomment-2187457396, depends on PR https://github.com/lopsided98/nix-ros-overlay/pull/442

### DIFF
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -143,12 +143,45 @@ in with lib; {
     '';
   });
 
+  # This is a newer version than the build system tries to download,
+  # cleaned definition to fix build for macOS on ARM64 since v1.12.9
   rviz-ogre-vendor = (patchVendorUrl rosSuper.rviz-ogre-vendor {
-    url = "https://github.com/OGRECave/ogre/archive/v1.12.1.zip";
-    sha256 = "1iv6k0dwdzg5nnzw2mcgcl663q4f7p2kj7nhs8afnsikrzxxgsi4";
-  }).overrideAttrs ({ ... }: {
+    originalUrl = "https://github.com/OGRECave/ogre/archive/v1.12.1.zip";
+    url = "https://github.com/OGRECave/ogre/archive/v1.12.10.zip";
+    hash = "sha256-lZDLywgShlWeWah7oTnyKBTqzN505LJKbQbgXRfJXlk=";
+  }).overrideAttrs ({
+    propagatedBuildInputs ? [], nativeBuildInputs ? [], patches ? [], prePatch ? "", postPatch ? "", ...
+  }: {
     # Prevent replacing $out/opt/.. with $out/var/empty/..
     dontFixCmake = true;
+    propagatedBuildInputs = propagatedBuildInputs ++ [ self.pugixml self.glew ];
+    nativeBuildInputs = nativeBuildInputs ++ lib.optionals self.stdenv.isDarwin [
+      self.darwin.apple_sdk.frameworks.Foundation
+      self.darwin.apple_sdk.frameworks.AppKit
+    ];
+    patches = patches ++ [
+      # Fix build failures on macOS + Apple Silicon
+      (self.fetchpatch {
+        url = "https://github.com/ros2/rviz/commit/0ef2b56373b98b5536f0f817c11dc2b5549f391d.patch";
+        hash = "sha256-tS8UGHih29T8AKCe+/ZEEEYDKNyY3bZRR/0nhiWaXlU=";
+        stripLen = 2;
+        extraPrefix = "";
+        postFetch = ''
+          substituteInPlace $downloadedFile \
+          --replace-fail '@@ -191,7 +195,7 @@' '@@ -188,7 +192,7 @@' \
+          --replace-fail 'relocatable.patch' 'fix-arm64.diff' \
+          --replace-fail 'C4267.patch' 'relocatable.patch'
+        '';
+      })
+    ];
+    prePatch = prePatch + ''
+      echo -n > fix-arm64.diff
+    '';
+    postPatch = postPatch + ''
+      substituteInPlace CMakeLists.txt \
+      --replace-fail 'ogre-v1.12.1' 'ogre-v1.12.10' \
+      --replace-fail 'URL_MD5 cdbea4006d223c173e0a93864111b936' "URL_MD5 c1b870955efddf539385094e9034e7f7"
+    '';
   });
 
   shared-queues-vendor = patchVendorUrl (patchVendorUrl rosSuper.shared-queues-vendor {
--- a/distros/ros2-overlay.nix
+++ b/distros/ros2-overlay.nix
@@ -215,7 +215,7 @@ rosSelf: rosSuper: with rosSelf.lib; {
   }: {
     dontWrapQtApps = false;
     nativeBuildInputs = nativeBuildInputs ++ [ self.qt5.wrapQtAppsHook ];
-    qtWrapperArgs = qtWrapperArgs ++ [
+    qtWrapperArgs = qtWrapperArgs ++ optionals self.stdenv.isLinux [
       # Use X11 by default in RViz2.
       # https://github.com/ros-visualization/rviz/issues/1442
       "--set-default QT_QPA_PLATFORM xcb"
--- a/pkgs/colcon/cmake.nix
+++ b/pkgs/colcon/cmake.nix
@@ -3,11 +3,11 @@
 
 buildPythonPackage rec {
   pname = "colcon-cmake";
-  version = "0.2.28";
+  version = "0.2.29";
 
   src = fetchPypi {
     inherit pname version;
-    hash = "sha256-XpEIkBSNnTx/gvXA6XkER44OC/FWt4ZykJnq1z1uxPA=";
+    hash = "sha256-8cChTiUw07c4+NBlnCfVnioKA9/vYB5hNpu1CD6/H2k=";
   };
 
   postPatch = ''
--- a/pkgs/colcon/core.nix
+++ b/pkgs/colcon/core.nix
@@ -29,11 +29,11 @@ let
 
   package = buildPythonPackage rec {
     pname = "colcon-core";
-    version = "0.18.3";
+    version = "0.18.4";
 
     src = fetchPypi {
       inherit pname version;
-      hash = "sha256-iz3PCxQLKFGpFpxgs+0sp4puSP7I7Oze76KHpnFfLhc=";
+      hash = "sha256-Oewm3WbVxppfQDEjnJMlgGn2H6Pfb4h0SjcAc9JpGh0=";
     };
 
     propagatedBuildInputs = [
--- a/pkgs/colcon/defaults.nix
+++ b/pkgs/colcon/defaults.nix
@@ -2,11 +2,11 @@
 
 buildPythonPackage rec {
   pname = "colcon-defaults";
-  version = "0.2.8";
+  version = "0.2.9";
 
   src = fetchPypi {
     inherit pname version;
-    hash = "sha256-BT6KGPuuBM8YKilo1/ftR0xRJb87MGuASSUFdPQJb6E=";
+    hash = "sha256-sPPMkAYmkGpGrp7lNB94sjeQjflXKhDlfQ8EHZRriOM=";
   };
 
   propagatedBuildInputs = [


etherswangel:
Only humble modified and tested by visualizing pointcloud2 played from a ros2bag. Needs test on linux and other macOS versions, also on other ros2 distros. @wentasah @hegde-atri could you help please?

distros/humble/rviz-ogre-vendor/rviz-ogre-vendor.patch
```diff
@@ -0,0 +1,55 @@
...
++  ExternalProject_Add(ogre-v1.12.13
++    URL https://github.com/OGRECave/ogre/archive/v1.12.13.zip
+     TIMEOUT 1200
+     LOG_CONFIGURE ${should_log}
+     LOG_BUILD ${should_log}
+@@ -185,21 +188,21 @@ macro(build_ogre)
+       -DOGRE_BUILD_COMPONENT_JAVA:BOOL=FALSE
+       -DOGRE_BUILD_COMPONENT_CSHARP:BOOL=FALSE
+       -DOGRE_BUILD_COMPONENT_BITES:BOOL=FALSE
++      -DOGRE_ENABLE_PRECOMPILED_HEADERS=0
```
wentasah, line 34: This seems to be the actual needed change besides the version bump. If this is the case, can you mention which error is actually fixed by this in the commit message?

distros/humble/rviz-ogre-vendor/rviz-ogre-vendor.patch
```diff
@@ -0,0 +1,55 @@
...
+       ${extra_cmake_args}
+       -Wno-dev
+     PATCH_COMMAND
+       ${Patch_EXECUTABLE} -p1 -N < ${CMAKE_CURRENT_SOURCE_DIR}/pragma-patch.diff &&
+-      ${Patch_EXECUTABLE} -p1 -N < ${CMAKE_CURRENT_SOURCE_DIR}/fix-arm64.diff &&
```
wentasah, line 39: I wounder whether this and the above change could also be made with `substituteInPlace`.

distros/humble/rviz-rendering/default.nix
```diff
@@ -2,7 +2,7 @@
 # Copyright 2024 Open Source Robotics Foundation
 # Distributed under the terms of the BSD license
 
-{ lib, buildRosPackage, fetchurl, ament-cmake, ament-cmake-cppcheck, ament-cmake-cpplint, ament-cmake-gmock, ament-cmake-gtest, ament-cmake-lint-cmake, ament-cmake-uncrustify, ament-cmake-xmllint, ament-index-cpp, ament-lint-auto, eigen, eigen3-cmake-module, qt5, resource-retriever, rviz-assimp-vendor, rviz-ogre-vendor }:
+{ lib, buildRosPackage, fetchurl, ament-cmake, ament-cmake-cppcheck, ament-cmake-cpplint, ament-cmake-gmock, ament-cmake-gtest, ament-cmake-lint-cmake, ament-cmake-uncrustify, ament-cmake-xmllint, ament-index-cpp, ament-lint-auto, eigen, eigen3-cmake-module, qt5, resource-retriever, rviz-assimp-vendor, rviz-ogre-vendor, glew }:
```
wentasah, line 5: This should be also in `overrides.nix`.

examples/ros2-basic.nix
```diff
@@ -8,9 +8,11 @@ mkShell {
   nativeBuildInputs = [
     (buildEnv {
       paths = [
-        ros-core
         colcon
-        geometry-msgs
+        ros-core
+        rviz2
+        ros2bag
+        sensor-msgs
```
wentasah, line 15: I think we can update the example, but I'd do it in a separate commit with an explanation why.

And I'm not sure whether we want to include `rviz2` in the basic example. rviz seems to be broken quite often so I'd prefer to have a simple example with CLI tools only, which demonstrates basic functionality and works most of the time. Then, we can add a more elaborate example, say `ros2-desktop`, which would contain the [dekstop](https://www.ros.org/reps/rep-2001.html#id25) entry point.

wentasah (changes requested):
Some changes will be needed. See the comments below. Also, when I build `rviz2` on Linux, I'm getting the following error:
```
-- Configuring done (8.7s)
CMake Error at Components/Overlay/CMakeLists.txt:42 (add_library):
  Cannot find source file:

    /build/rviz-release-release-humble-rviz_ogre_vendor-11.2.12-1/build/ogre-v1.12.13-prefix/src/ogre-v1.12.13-build/imgui-1.79/imgui.cpp
```

examples/ros2-basic.nix
etherswangel: Sorry it's a mistake, forgot to remove that after testing, will revert that

distros/humble/rviz-ogre-vendor/rviz-ogre-vendor.patch
```diff
@@ -0,0 +1,55 @@
...
+     # include Clang -Wno-everything to disable warnings in that build. GCC doesn't mind it
+@@ -161,9 +165,8 @@ macro(build_ogre)
+ 
+   find_package(Patch REQUIRED)
+   include(ExternalProject)
+-  ExternalProject_Add(ogre-v1.12.1
+-    URL https://github.com/OGRECave/ogre/archive/v1.12.1.zip
+-    URL_MD5 cdbea4006d223c173e0a93864111b936
++  ExternalProject_Add(ogre-v1.12.13
++    URL https://github.com/OGRECave/ogre/archive/v1.12.13.zip
```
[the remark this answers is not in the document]
etherswangel: resolved but I'm not sure how 'URL_MD5' should be handled, currently replaced by `substituteInPlace`

distros/humble/rviz-ogre-vendor/rviz-ogre-vendor.patch
etherswangel: Without this line we get `error: Objective-C was disabled in PCH file but is currently enabled`. Probably because the upstream don't provide precompiled headers for macOS.

etherswangel:
Thank you @wentasah! patch for `darwin-aarch64` is pushed to https://github.com/ros2-gbp/rviz-release/commit/777d3f45f7d6197e3ee3b7e3b9e95f0138ca6266, version bump in done by `substituteInPlace`. 

However I don't have Linux machine, can you provide more outputs? I think I've encountered this before in other repos, will do my best to help.

distros/humble/overrides.nix
```diff
@@ -130,10 +130,33 @@ in with lib; {
...
+  }: {
+    propagatedBuildInputs = propagatedBuildInputs ++ [ self.pugixml self.glew ];
+    nativeBuildInputs = nativeBuildInputs ++ lib.optionals self.stdenv.isDarwin [
+      self.darwin.apple_sdk.frameworks.Foundation
+      self.darwin.apple_sdk.frameworks.AppKit
+    ];
+    patches = patches ++ lib.optionals self.stdenv.isDarwin [
+      # Fix build with darwin-aarch64
+      (self.fetchpatch {
+        url = "https://github.com/ros2-gbp/rviz-release/commit/777d3f45f7d6197e3ee3b7e3b9e95f0138ca6266.patch";
```
wentasah, line 150: This is not what I meant. AFAICS the commit you're referring to is only in [your repository](https://github.com/etherswangel/rviz-release/tree/patch/darwin-aarch64). If it's needed, you should at least use the URL of your repo. However, it's not needed in this case, because the same fixes are already merged in the upstream repo (e.g. [here](https://github.com/ros2/rviz/pull/944/files) and [here](https://github.com/ros2/rviz/pull/1156/files)). The preferred way is to use those patches.

wentasah (changes requested):
Just a quick note for now. I'll investigate the build failure on Linux later when I have more time. Maybe, it will go away if you use the upstream patches.

etherswangel:
Hi there! Any updates during this week? 

In summary,
- the building failure on Linux. Is there any progress?
- are `pugixml` and `glew` needed on other platforms?
- should I change the URL to my repo or try to find a upstream patch that can be used directly?

I’ll definitely set up a Linux container for future contributions when I have some free time. But for now, I’m afraid I can’t help with this.

hegde-atri:
Sorry, I got swamped with other work and forgot about this. I will have a go at tackling the building failure on Linux tomorrow and let you know how I get along.

hegde-atri:
Currently getting a failed patch error while building rviz-ogre-vendor on linux. I will have more time to work on this tomorrow. As for your other points (2 & 3), I have no clue 😅

hegde-atri:
@wentasah can you give me any pointers for the issue on linux? Currently its failing while building `rviz-ogre-vendor`.
```
Hunk #2 succeeded at 112 with fuzz 2 (offset -1 lines).
(Stripping trailing CRs from patch; use --binary to disable.)
patching file PlugIns/OctreeZone/CMakeLists.txt
(Stripping trailing CRs from patch; use --binary to disable.)
patching file OgreMain/include/OgreTimer.h
patching file OgreMain/src/OgrePlatformInformation.cpp
Reversed (or previously applied) patch detected!  Skipping patch.
1 out of 1 hunk ignored -- saving rejects to file OgreMain/src/OgrePlatformInformation.cpp.rej
make[2]: *** [CMakeFiles/ogre-v1.12.13.dir/build.make:115: ogre-v1.12.13-prefix/src/ogre-v1.12.13-stamp/ogre-v1.12.13-patch] Error 1
make[1]: *** [CMakeFiles/Makefile2:137: CMakeFiles/ogre-v1.12.13.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

stevalkr:
Hello, PR #442 is merged, so I rebased this branch and fixed https://github.com/lopsided98/nix-ros-overlay/pull/445#issuecomment-2322898732 @hegde-atri. However I don't have the issue https://github.com/lopsided98/nix-ros-overlay/pull/445#pullrequestreview-2217194085 @wentasah. Tested on macOS 15.1 and ubuntu 24.04 docker container.

Could you please take a look at this PR and let me know if you spot any other issues? If everything looks good enough, I’ll go ahead and squash it and then wait for a merge. Thanks a bunch!

lopsided98 (changes requested):
Sorry for neglecting this, but unfortunately I'm not going to merge this for now. My policy is that if upstream does not support macOS, then I don't want to maintain a bunch of hacks to make it work. I have no ability to test on macOS, so complex patches hinder my ability to maintain the overlay for Linux.

It does mainly look like the problems are with aarch64-darwin, so I would be willing to merge a more limited PR that only targets x86_64-darwin.

lopsided98:
If there are upstream patches in recent distros that fix the problem, backporting them to humble is fine. I just want to things to start trending toward getting fixed upstream, so I don't have to maintain downstream patches forever.

lopsided98:
I would also appreciate it if you could try to get newer distros working as well. In particular, the macOS dependency overrides probably be moved to `ros2-overlay.nix`.

stevalkr:
@lopsided98 thanks for reaching out, I’ve got a fix but it’s a bit of a hack. Please take a look and let me know if it works for you!

lopsided98:
Sorry, I didn't mean to make you add a bunch of hacks just to avoid making a custom version of the patch. I also didn't look at it closely enough before to realize how similar your original patch was to the version accepted upstream.

On the other hand, is it actually okay to just replace Ogre 1.12.1 with 1.12.10? The upstream commit that upgraded Ogre (https://github.com/ros2/rviz/commit/4e976dc444e7e59ec9cdef30f1f04e0beafcafc8) looks to have some non-trivial changes to make it compatible. It might build, but does it run properly?

I assume you are stuck on humble. It would be a lot easier if we could just forget about humble and make newer distros work instead...

stevalkr:
I haven’t had a chance to fully test it, but it seems to work on my Mac (MacOS 15.1.1, M3 Max). I’m not sure if it breaks any functionalities I don’t use or Linux, though.

It looks like newer distributions have higher versions of all the dependencies, so the patches won’t apply to all of them right away. Since humble is still the latest LTS, I was planning to work on iron after the rviz2 and rqt pull requests are merged. Nevertheless, I can begin working on iron later. I’ll start a new pull request then.

lopsided98:
It looks like both Iron and Jazzy have been upgraded to Ogre 1.12.10 and have https://github.com/ros2/rviz/commit/0ef2b56373b98b5536f0f817c11dc2b5549f391d applied. What is there to patch other than adding the macOS framework dependencies?

wentasah:
Closing. See #930.

Issue #460, opened 2024-08-31
RQT (ROS2 Humble) fails to build on MacOS (arm64)


hegde-atri:
Adding `rqt` to my package list, it fails to build.

First there is an error about `shiboken2` being marked as broken.
`error: Package ‘shiboken2-5.15.11’ in /nix/store/snp5f95xhkz89kdw9703kana3m6xfd0a-source/pkgs/development/python-modules/shiboken2/default.nix:54 is marked as broken, refusing to evaluate.`

After overriding this, `tango-icon-theme` fails to build is it is not available on requested platform.

After also overriding this, I get a build error with gnome-icon-theme.

```
error: builder for '/nix/store/irz9n8mhrcdkbg5g3qb1jb0xmb424hfg-gnome-icon-theme-3.12.0.drv' failed with exit code 1;
       last 25 log lines:
       > for lang in $linguas; do \
       >   rm -f /nix/store/0mn1gp5s1czx5dpz18y2is3ij6k3sp6n-gnome-icon-theme-3.12.0/lib/locale/$lang/LC_MESSAGES/gnome-icon-theme.mo; \
       >   rm -f /nix/store/0mn1gp5s1czx5dpz18y2is3ij6k3sp6n-gnome-icon-theme-3.12.0/lib/locale/$lang/LC_MESSAGES/gnome-icon-theme.mo.m; \
       > done
       > make[3]: Leaving directory '/private/tmp/nix-build-gnome-icon-theme-3.12.0.drv-0/gnome-icon-theme-3.12.0/po'
       > if [ -h /nix/store/0mn1gp5s1czx5dpz18y2is3ij6k3sp6n-gnome-icon-theme-3.12.0/share/icons/gnome ]; then \
       >   rm -f /nix/store/0mn1gp5s1czx5dpz18y2is3ij6k3sp6n-gnome-icon-theme-3.12.0/share/icons/gnome; \
       > fi
       > if [ -d /nix/store/0mn1gp5s1czx5dpz18y2is3ij6k3sp6n-gnome-icon-theme-3.12.0/share/icons/gnome ]; then \
       >    touch /nix/store/0mn1gp5s1czx5dpz18y2is3ij6k3sp6n-gnome-icon-theme-3.12.0/share/icons/gnome; \
       > fi
[...]

PR #471, opened 2024-09-15
fix: remove tango-icons-vendor and pyside2 for macos

stevalkr:
fix issue 
- #460 

tested on macOS 15.1 and ubuntu 24.04 container


hegde-atri:
This seems to fix the issue for me! (macOS 15.1)

wentasah:
@stevalkr Could you add a short comment to each MacOS exception explaining why it is needed? I can imagine that sooner or later somebody figures out that this dependency is needed to make some other package work and they will afraid to remove the exception in the fear of breaking something else. I thinks that something like `# Needed for PackageXY build to succeed` would be sufficient.

stevalkr:
I suppose these are you mean @wentasah? Well my English isn't good, please edit if they look strange 😂

wentasah:
I mean, you probably didn't want to install `tango-icons-vendor` or `python-qt-binding`, but some other package that depends on these. Was is `rqt`? Or something else? Mention that package in the comment.

PR #422, opened 2024-07-06
Patch gz-*-vendor packages for Jazzy and Rolling

wentasah:
This is my (incomplete) attempt to compile `rviz2` for Jazzy. It requires patching vendored packages, which is what this PR does. However, it's not sufficient - I get the following error:

```
-- Found gz_math_vendor: 0.0.5 (/nix/store/zncdwq643ism55hpyxlb8rg2zxfcddbz-ros-jazzy-gz-math-vendor-0.0.5-r1/share/gz_math_vendor/cmake)
CMake Error at CMakeLists.txt:65 (find_package):
  By not providing "Findgz-math.cmake" in CMAKE_MODULE_PATH this project has
```

Probably, there is some logic in `ament_vendor`, which we do not patch correctly. It will need more investigation. I'll be mostly off-line next week so if someone wants to continue, they can start here.

PR #472, opened 2024-09-15
fix: gz vendor

muellerbernd:
this work builds on top of #422 and now `gz sim` works in both jazzy and rolling. I have also included a dev shell for gazebo development.

I had to add :
```nix
preBuild = ''
      find . -name "*build.make" -print -exec sed -i "s#var/empty#opt#g" {} \; '';
```
to the `patch*AmentVendor` functions, because even with `dontFixCmake=true` the `CmakeConfigureHook` replaced `/opt` with `/var/empty` in CMakeInstallPrefix of the `*build.make` file. 

This also works with local version of `ros_gz`, but I had to change the launch file (ros_gz/ros_gz_sim/launch/gz_sim.launch.py.in) in such a way that it does not use `ruby $(which gz) sim` to start the simulator. 

`ruby $(which gz) sim` causes the following errors:
```
/nix/store/5yv2ncm0iwwwbvad44znjvdrq7x8ass1-ros-env/opt/gz_tools_vendor/bin/gz:1: Invalid char `\x7F' in expression
/nix/store/5yv2ncm0iwwwbvad44znjvdrq7x8ass1-ros-env/opt/gz_tools_vendor/bin/gz:1: Invalid char `\x02' in expression
/nix/store/5yv2ncm0iwwwbvad44znjvdrq7x8ass1-ros-env/opt/gz_tools_vendor/bin/gz:1: Invalid char `\x01' in expression
/nix/store/5yv2ncm0iwwwbvad44znjvdrq7x8ass1-ros-env/opt/gz_tools_vendor/bin/gz:1: Invalid char `\x01' in expression
```


wentasah (changes requested):
This looks promising (but one has to rebase it onto #422 and look at diff against that).

I have a few comments or suggestions (see below).

Also, I'm not sure, it works for me (I normally don't use gazebo). When I run `gz sim`, I see the title screen. I select an image and press "RUN". The title screen disappears and nothing happens. Is it expected? Shall I wait longer time? The CPU load is almost nothing.

muellerbernd:
For testing a simple simulation you can use:  
```bash
gz sim shapes.sdf 
```
This will download some files for the shape models and will start a simulation with different shapes loaded. Here you can find [more documentation](https://gazebosim.org/docs/latest/getstarted/)

https://github.com/user-attachments/assets/2ee6f30a-9469-4ed0-8f1f-53866d1869e0

hacker1024:
Thanks for this! I've added some changes to make `gz` work outside of `nix-shell` as well: https://github.com/muellerbernd/nix-ros-overlay/pull/2.

lopsided98:
I merged this manually and added some patching of ament-cmake-vendor-package to make it install to normal directories instead of `/opt`. This avoids all the special handling for Gazebo packages in `buildEnv` and doesn't appear to break anything.

PR #496, opened 2024-09-30
[fix] replace distutils.spawn in sip_configure.py

muellerbernd:
Replace distutils.spawn in sip_configure.py if present. For jazzy this is fixed upstream but for humble not.

### DIFF
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -206,6 +206,7 @@ let
             -e "s#with open(os.path.join(tmpdirname, 'QtCore', 'QtCoremod.sip'), 'w') as outfp:##" \
             -e "s#outfp.write(output)##" \
             -i cmake/sip_configure.py
+        substituteInPlace cmake/sip_configure.py --replace "from distutils.spawn import find_executable" "from shutil import which as find_executable"
       '';
     });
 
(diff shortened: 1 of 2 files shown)


wentasah:
What is this supposed to fix? Building `rosPackages.humble.python-qt-binding` works even without this PR.

Wouldn't it be better to apply the upstream fix? I guess it's [this one](https://github.com/ros-visualization/python_qt_binding/commit/ee4d43bcdb0c5c5d40f81dea3de6185298ab34a7).

muellerbernd:
This fixes building of `qt-gui-cpp`. Thanks for pointing out the upstream fix. I have updated the PR.

wentasah:
This looks better. Probably, the same fix would be needed for noetic and iron, which also fail to build.

I wonder what @lopsided98 prefers. Whether repeating the same code in multiple `distros/*/overrides.nix` (which was being done in the past) or having it once in in `distro-overlay.nix` applied conditionally based on the value of `distro` parameter (which is enabled by a recent patch). I'd say the latter is now preferable.

lopsided98:
I actually prefer copying the code between distros. That way patches for old distros don't stick around after the distro is deleted.

Issue #491, opened 2024-09-24
Since b6676bf cmake_modules cannot be found


kjeremy:
Packages can't find  `cmake-modules` anymore. On [dde852a](https://github.com/lopsided98/nix-ros-overlay/commit/dde852afc9f761d150950d44c8b03ee5a19dc73e)

Bisected to b6676bf7dc1b03afe9582c5fd99025b3a4b7f4f9

```console
[jkolb@nixos:~/nix-ros-overlay]$ nix build .#noetic.pluginlib
error: builder for '/nix/store/rqnrly7cjmxylva3fypjn9dq6blbny3r-ros-noetic-class-loader-0.5.0-r1.drv' failed with exit code 1;
       last 25 log lines:
       > -- Using PYTHON_EXECUTABLE: /nix/store/h3i0acpmr8mrjx07519xxmidv8mpax4y-python3-3.12.5/bin/python
       > -- Using default Python package layout
       > -- Found PY_em: /nix/store/293x651b1sv2w9nvi8mxh5av2rcyp18j-python3.12-empy-4.1/lib/python3.12/site-packages/em.py
       > -- Using empy: /nix/store/293x651b1sv2w9nvi8mxh5av2rcyp18j-python3.12-empy-4.1/lib/python3.12/site-packages/em.py
       > -- Using CATKIN_ENABLE_TESTING: OFF
       > -- catkin 0.8.10
       > -- BUILD_SHARED_LIBS is on
       > -- Could NOT find cmake_modules (missing: cmake_modules_DIR)
       > -- Could not find the required component 'cmake_modules'. The following CMake error indicates that you either need to install the package with the same name or change your environment so that it can be found.
       > CMake Error at /nix/store/d6y7v13kik3kl2a673qxpkw0hsm56zxs-ros-noetic-catkin-0.8.10-r1/share/catkin/cmake/catkinConfig.cmake:83 (find_package):
       >   Could not find a package configuration file provided by "cmake_modules"
       >   with any of the following names:
       >
       >     cmake_modulesConfig.cmake
       >     cmake_modules-config.cmake
       >
       >   Add the installation prefix of "cmake_modules" to CMAKE_PREFIX_PATH or set
       >   "cmake_modules_DIR" to a directory containing one of the above files.  If
       >   "cmake_modules" provides a separate development package or SDK, be sure it
       >   has been installed.
       > Call Stack (most recent call first):
       >   CMakeLists.txt:22 (find_package)
       >
       > 
[...]

PR #497, opened 2024-10-02
export CMAKE_PREFIX_PATH from the catkin setup hook

kjeremy:
Fixes: #491

### DIFF
--- a/distros/catkin-setup-hook/setup-hook.sh
+++ b/distros/catkin-setup-hook/setup-hook.sh
@@ -73,15 +73,15 @@ _catkinPreConfigureHook() {
   # we still need the marker for the environment hooks, so it is created
   # manually.
   cmakeFlags+=" -DCATKIN_BUILD_BINARY_PACKAGE=ON"
+
+  # Catkin uses CMAKE_PREFIX_PATH to find additional workspaces, but
+  # nixpkgs switched to using NIXPKGS_CMAKE_PREFIX_PATH
+  # (https://github.com/NixOS/nixpkgs/commit/8c9c8ade2f88a85ccdd4858cc802d7b7d6c48fe0).
+  # see: https://github.com/lopsided98/nix-ros-overlay/issues/491
+  # When specified as a variable CMAKE_PREFIX_PATH is a semi-colon separated list.
+  cmakeFlags+=" -DCMAKE_PREFIX_PATH=${NIXPKGS_CMAKE_PREFIX_PATH//:/;}"
 }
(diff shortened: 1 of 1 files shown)


distros/catkin-setup-hook/setup-hook.sh
```diff
@@ -76,6 +76,13 @@ _catkinPreConfigureHook() {
 }
 preConfigureHooks+=(_catkinPreConfigureHook)
 
+_exportCMakePrefixPath() {
+  # Needed for https://github.com/NixOS/nixpkgs/commit/8c9c8ade2f88a85ccdd4858cc802d7b7d6c48fe0
+  # see: https://github.com/lopsided98/nix-ros-overlay/issues/491
+  export CMAKE_PREFIX_PATH=$NIXPKGS_CMAKE_PREFIX_PATH
```
wentasah, line 82: I'd just add this to `_catkinPreConfigureHook`. It's not necessary to create a new function.

And the comment could mention why it's needed - I propose something like this:

```
Catkin uses CMAKE_PREFIX_PATH to find additional workspaces, but
nixpkgs switched to using NIXPKGS_CMAKE_PREFIX_PATH
(https://github.com/NixOS/nixpkgs/commit/8c9c8ade2f88a85ccdd4858cc802d7b7d6c48fe0).
see: https://github.com/lopsided98/nix-ros-overlay/issues/491
```

I also wonder whether it's necessary to export the variable to all processes running during build. It seems it should be sufficient to pass it just to cmake via command line switches:

```sh
cmakeFlags+=" -DCMAKE_PREFIX_PATH=\"$NIXPKGS_CMAKE_PREFIX_PATH\""
```

Can you check, whether this works? If yes, I'd prefer this variant, because it might help with some cross-compilation scenarios, which are fixed by the nixpkgs commit.

wentasah (changes requested):

wentasah (approved):
LGTM

kjeremy:
This doesn't solve the issue in devshells though.

wentasah:
I guess that `cmake $cmakeFlags ..` should work. Does it?

If you want it to work without `$cmakeFlags`, we'll need to patch those catkin cmake files that use `CMAKE_PREFIX_PATH` to also use `NIXPKGS_CMAKE_PREFIX_PATH`.

kjeremy:
> I guess that `cmake $cmakeFlags ..` should work. Does it?

No. `$cmakeFlags` is empty.

kjeremy:
Closing in favor of #500

Issue #511, opened 2024-10-20
velodyne-pointcloud 2.4.0-r1 doesn't build


quantum9Innovation:
See: https://github.com/lopsided98/nix-ros-overlay/blob/master/distros/humble/velodyne-pointcloud/default.nix

## Steps to reproduce

1. Load the following into `flake.nix`:

```nix
{
  inputs = {
    nixpkgs.follows = "nix-ros-overlay/nixpkgs";
    nix-ros-overlay.url = "github:lopsided98/nix-ros-overlay/master";
  };
  outputs =
    {
      self,
      nix-ros-overlay,
      nixpkgs,
    }:
    nix-ros-overlay.inputs.flake-utils.lib.eachDefaultSystem (
      system:
      let
        pkgs = import nixpkgs {
          inherit system;
          overlays = [ nix-ros-overlay.overlays.default ];
        };
      in
      {
        devShells.default = pkgs.mkShell {
          name = "Test";
          packages = [
            (
              with pkgs.rosPackages.humble;
              buildEnv {
                paths = [
   	              ros-core
	              velodyne-pointcloud
                  # add other ROS packages here
                ];
              }
            )
          ];
        };
      }
    );
  nixConfig = {
    extra-substituters = [ "https://ros.cachix.org" ];
    extra-trusted-public-keys = [ "ros.cachix.org-1:dSyZxI8geDCJrwgvCOHDoAfOm5sV1wCPjBkKL+38Rvo=" ];
  };
}
```

2. `nix develop` gives:

```
error: builder for '/nix/store/qv2kmym5qwp6mfs2075sjr7ca1ypqbrj-ros-humble-velodyne-pointcloud-2.4.0-r1.drv' failed with exit code 1;
       last 25 log lines:
       > -- Found rmw_fastrtps_cpp: 6.2.6 (/nix/store/ma7myf6hai606m5iijwwivid92f4jaxi-ros-humble-rmw-fastrtps-cpp-6.2.6-r1/share/rmw_fastrtps_cpp/cmake)
       > -- Found OpenSSL: /nix/store/1l3a02nqq5b5v7rhchj89hi7plmbza5r-openssl-3.0.12/lib/libcrypto.so (found version "3.0.12")
       > -- Found FastRTPS: /nix/store/vp25xly0qsyvkimi1hmcjhi9fb4rnfk7-ros-humble-fastrtps-2.6.7-r1/include
[...]

PR #514, opened 2024-10-22
moveit2: fix build for Jazzy

okvik:
This fixes the build of MoveIt2 for Jazzy. The patches likely apply equally for Rolling but I have no time to test this currently.

It's been a while since I did this so exact details escape me, and again, I lack time to re-research, but in short:

1) Several packages have constrained their version requirement on `liboctomap` to <1.10 to fix some distro specific issues on Ubuntu, while what we have is =1.10. As far as I understand the issue is irrelevant on Nix. The fix is to relax the version requirement.
2) `osqp_vendor` required a bit of convincing to get it to build and install like we want.

### DIFF
--- a/distros/jazzy/overrides.nix
+++ b/distros/jazzy/overrides.nix
@@ -25,14 +25,6 @@ in {
 
   gazebo = self.gazebo_11;
 
-  geometric-shapes = rosSuper.geometric-shapes.overrideAttrs({
-    postPatch ? "", ...
-  }: {
-    postPatch = postPatch + ''
-      substituteInPlace CMakeLists.txt --replace-fail 'find_package(octomap 1.9.7...<1.10.0 REQUIRED)' 'find_package(octomap REQUIRED)'
-    '';
-  });
-
   google-benchmark-vendor = lib.patchExternalProjectGit rosSuper.google-benchmark-vendor {
     url = "https://github.com/google/benchmark.git";
     rev = "344117638c8ff7e239044fd0fa7085839fc03021";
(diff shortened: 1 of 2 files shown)


distros/jazzy/overrides.nix
```diff
@@ -184,6 +192,51 @@ in {
...
+    postPatch = postPatch + ''
+      substituteInPlace CMakeLists.txt --replace-fail 'find_package(octomap 1.9.7...<1.10.0 REQUIRED)' 'find_package(octomap REQUIRED)'
+    '';
+  });
+
+  osqp-vendor = lib.pipe rosSuper.osqp-vendor [
+    # Make CMakeLists.txt amenable to automatic patching by the next step.
+    (pkg: pkg.overrideAttrs ({ prePatch ? "", ... }: {
+      prePatch = prePatch + ''
+        sed -i 's/GIT_TAG.*/GIT_TAG v0.6.2/' CMakeLists.txt
```
wentasah, line 215: What about being more explicit? Something like this:
```suggestion
        substituteInPlace CMakeLists.txt --replace-fail 'set(git_tag "v0.6.2")' 'set(git_tag "v0.6.2")' # fail when upstream version changes
        substituteInPlace CMakeLists.txt --replace-fail 'GIT_TAG ''${git_tag}' 'GIT_TAG v0.6.2'
```

wentasah (changes requested):
The same changes seem to be needed for rolling. What about placing it in `distro-overlay.nix` so that it can be applied for both versions, similarly as in https://github.com/lopsided98/nix-ros-overlay/issues/511#issuecomment-2425252050.

distros/jazzy/overrides.nix
okvik: Sounds good.

okvik:
> The same changes seem to be needed for rolling. What about placing it in `distro-overlay.nix` so that it can be applied for both versions, similarly as in [#511 (comment)](https://github.com/lopsided98/nix-ros-overlay/issues/511#issuecomment-2425252050).

It surely applies to Rolling, at least currently. It's just unclear to me what we want to happen to Rolling (and future distros) when the patches inevitably fail to apply (due to updates to versions); also what versions and other things might be needed for older distros. None of this I had time or interest in checking because I don't use these, so it just didn't seem fair of me to assume. But looking at it from another point of view, we can let things break and fix when they actually come up.

Issue #264, opened 2023-05-18
MoveIt2 move_group, RViz2, etc. can't find class plugins


okvik:
After some patching and cherry picking issues I almost have a workable ROS 2 Humble environment with MoveIt2 and RViz2.

However, the `move_group` package that expects to load its capabilities from plugins have trouble finding them, even though the plugin objects themselves and the resource index stuff seem to be in place. Strangely enough the MoveIt2 planning pipelines, which are also plugins, get loaded fine... 

I'm not exactly sure on how to continue debugging this. Did anyone encounter this with other packages before? I can't seem to find many references to relevant problems with plugins in past issues.

Later today I'll (try to) provide a way to reproduce this easily, but for now here's a log of running `ros2 launch moveit_resources_fanuc_moveit_config demo.launch.py` (Fanuc instead of Panda here because Panda config package is broken in the Humble release of `MoveIt`... 

```
[INFO] [launch]: All log files can be found below /home/kvik/.ros/log/2023-05-18-16-22-50-727021-vpcd-dell-2379303
[INFO] [launch]: Default logging verbosity is set to INFO
Using load_yaml() directly is deprecated. Use xacro.load_yaml() instead.
[INFO] [rviz2-1]: process started with pid [2379304]
[INFO] [static_transform_publisher-2]: process started with pid [2379306]
[INFO] [robot_state_publisher-3]: process started with pid [2379308]
[INFO] [move_group-4]: process started with pid [2379310]
[INFO] [ros2_control_node-5]: process started with pid [2379312]
[INFO] [spawner-6]: process started with pid [2379314]
[INFO] [spawner-7]: process started with pid [2379316]
[static_transform_publisher-2] [WARN] [1684419771.395976573] []: Old-style arguments are deprecated; see --help for new-style arguments
[ros2_control_node-5] Warning: class_loader.impl: SEVERE WARNING!!! A namespace collision has occurred with plugin factory for class mock_components::GenericSystem. New factory will OVERWRITE existing one. This situation occurs when libraries containing plugins are directly linked against an executable (the one running right now generating this message). Please separate plugins out into their own library or just don't link against the library and use either class_loader::ClassLoader/MultiLibraryClassLoader to open.
[ros2_control_node-5]          at line 253 in /nix/store/4is2qsw3j6an766xhbzik16zzhrf6sg9-ros-humble-class-loader-2.2.0-r3/include/class_loader/class_loader/class_loader_core.hpp
[...]

PR #516, opened 2024-10-23
moveit2: fix move_group plugins not found issue

okvik:
This fixes the long-standing issue discussed in https://github.com/lopsided98/nix-ros-overlay/issues/264

Full credits for this code go to @khaninger with a big thanks to @Takishima for finding and sharing the actual issue.

I initially tried to recreate the approach suggested by @Takishima where various `CMakeFiles.txt` get patched but was unsuccessful as this caused other build failures that I had trouble solving, ultimately deciding on a straightforward approach that was already working for me (and apparently others) in practice.

This depends on https://github.com/lopsided98/nix-ros-overlay/pull/514 so when that gets merged I can rebase.


kjeremy:
Should this be applied to all ros2 distros not just jazzy?

okvik:
> Should this be applied to all ros2 distros not just jazzy?

Likely. The issue exists from at least Humble and very likely the same fix would work.

okvik:
I mucked up with Git here. Will reopen later when I fix things.

PR #517, opened 2024-10-24
Fix usb-cam build

IamTheCarl:
This package was failing to build because it was missing pkg-config.
I have not tested it on any distros other than humble, but I can add it and see if they at least build.

### DIFF
--- a/distros/humble/usb-cam/default.nix
+++ b/distros/humble/usb-cam/default.nix
@@ -1,8 +1,7 @@
-
 # Copyright 2024 Open Source Robotics Foundation
 # Distributed under the terms of the BSD license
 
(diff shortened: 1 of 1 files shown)


wentasah:
Thanks for the fix, but please don't modify the automatically generated files. The best fix is to add pkg-config to the upsteam `package.xml`. If that's not possible or not fast enough, add an override to overrides.nix. If multiple distros need the fix, it's IMHO the best to add the override to distro-overlay.nix as in https://github.com/lopsided98/nix-ros-overlay/issues/511#issuecomment-2425252050.

IamTheCarl:
You're welcome, and thank you for taking the time to review it. If I had read the README more closely, I would have realized this was the wrong way to go about things. A header at the top of generated files stating "Do not modify, this file was auto generated by superflore" would have clued me in a little harder.

I'm not closing out this PR yet because I don't know if upstream will accept, reject, or ignore my PR. If the PR is not accepted, I will rework this PR to use the `distro-overlay.nix` alternative you suggested.

IamTheCarl:
Upstream has been fixed.
https://github.com/ros-drivers/usb_cam/pull/347

PR #518, opened 2024-10-27
bloom: init at 0.12.0

nim65s:
Hi,

This add bloom and its vcstools required dependency.

I was able to create a new release with it: https://github.com/ros/rosdistro/pull/43276


wentasah:
We're sitting with @nim65s at Nixcon and I'm testing it with him. It works for me.

PR #519, opened 2024-10-27
Extend release.nix and add a link to public Hydra

### LINKED ISSUE #510: Faster CI?
Hi @lopsided98 ,

if I understand correctly, you build all ROS packages on GitHub runners and aarch64-linux packages are built via qemu. This must take very long time. At our university, I've got a few powerful servers, whose power I could donate to building ROS packages. Specifically, I have an 80-core aarch64-linux machine and a 32-core x86_64-linux machine. Would you be interested in using these to speed-up development in this repo?

I see several possibilities how to use them. The simplest one, which can be working very quickly, is that the servers can be used as nix remote builders. Either for CI running on GitHub, or even for you personal Hydra (I think you have one somewhere).  Another option would be to run Hydra (or other CI system) on those servers to and integrate it in this repo.

Let me know what do you think and we can then discuss details such as security, trustworthiness etc.

wentasah:
I've setup [a public Hydra instance](https://hydra.iid.ciirc.cvut.cz/project/nix-ros-overlay
), which builds this overlay for x86 and aarch64 linux. Currently, I consider it experimental in the sense, that I might change the configuration in the future depending on how it will behave, but it already gives useful information for users and contributors.

To make it easier to setup Hydra jobsets, I updated release.nix. It's probably not worth reading all commits, one by one, because some later commits override previous ones and I was lazy to redo things cleanly from the beginning. But the final state is hopefully understandable.

Currently, the hydra builds my fork, but once this PR is merged, I'll switch it to the upstream repo. The configuration of the jobsets is generated programmatically and is available in [another repo](https://github.com/wentasah/nix-ros-hydra). 

Closes #510

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,9 @@ You can use the following command to use the above template easily:
 ## Current status
 
 What works:
-1. More than 1400 packages successfully built for ROS Noetic
+1. More than 1400 packages successfully built for ROS Noetic (for
+   up-to-date numbers and other distros, look at our experimental
+   [Hydra instance][] controlled by [@wentasah][])
 2. Fully functional ROS development environment using `nix-shell`
 3. Automated generation of Nix package definitions using standard ROS tools ([superflore](https://github.com/lopsided98/superflore))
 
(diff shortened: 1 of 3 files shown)


wentasah:
I've rebased this to develop and fixed the conflicts. Also added the last commit to not include `examples/flake` as the output of `release.nix`.

lopsided98:
Thanks for setting this up. I have my own Hydra instance, but I'm not comfortable making it publicly accessible.

One request is that you make it clear in the README that you control this instance. It's not too big of a deal since you aren't promoting it as a binary cache, but I think it is good to transparent about it.

wentasah:
I extended the readme as you suggested. I'll reply more to #510 later :-)

PR #534, opened 2024-12-06
Update meta.description and add meta.homepage if missing

Pandapip1:


pkgs/bloom/default.nix
```diff
@@ -35,7 +35,8 @@ buildPythonPackage rec {
   pythonImportsCheck = [ "bloom" ];
 
   meta = {
-    description = "A release automation tool which makes releasing catkin (http://ros.org/wiki/catkin) packages easier";
+    description = "A release automation tool which makes releasing catkin packages easier";
+    fullDescription = "A release automation tool which makes releasing catkin (http://ros.org/wiki/catkin) packages easier";
```
lopsided98, line 39: `fullDescription` doesn't exist, it's `longDescription`

pkgs/colcon/python-setup-py.nix
```diff
@@ -15,6 +15,7 @@ buildPythonPackage rec {
   doCheck = false;
 
   meta = with lib; {
+    description = "Extension for colcon-core to identify packages from setup.py files";
```
lopsided98, line 18: Duplicate attribute, the other needs to be renamed to `longDescription`.

pkgs/gazebo/gz-cmake/generic.nix
```diff
@@ -20,7 +20,7 @@ stdenv.mkDerivation rec {
 
   meta = with lib; {
     homepage = "https://gazebosim.org/libs/cmake";
-    description = "A set of CMake modules that are used by the C++-based Gazebo projects. ";
+    description = "A set of CMake modules that are used by the C++-based Gazebo projects";
```
lopsided98, line 23: Did you want to remove the indefinite article here like the rest?

lopsided98 (changes requested):
You need to evaluate all these packages, would have caught these issues and any others that I missed.

Pandapip1:
CI is failing on develop.

Issue #536, opened 2024-12-08
"ROS" packages already in nixpkgs


nim65s:
Hi,

Could we override some packages from the overlay to directly use those provided by nixpkgs, instead of trying to re-declare and re-build things from ROS ?

Context: I'm the maintainer (upstream + nixpkgs + ROS) of a few things (eg. eigenpy / hpp-fcl / pinocchio): pinocchio is in better shape in nixpkgs than in ROS. For example in nixpkgs we can enable the casadi optional dependency, while in ROS that's harder since casadi is not packaged in ubuntu.

Also, since pinocchio requires lot of RAM and time to build, I think it would be very beneficial for your build & cache infra :)

NB: I'm speaking of non-ROS-only packages, which are packaged in rosdistro. Maybe it make little or no difference for pure ROS packages, but version constraints might be harder.

nim65s:
Hum, wait, that's already the case, isn't it ?
You're wonderful, thank you.

nim65s:
Actually, that's half the case, this is why I was mislead in the first place:
```nix
nix repl
Lix 2.91.1
Type :? for help.
nix-repl> :lf github:lopsided98/nix-ros-overlay
Added 17 variables.

nix-repl> outputs.legacyPackages.x86_64-linux.humble.pinocchio                 
«derivation /nix/store/ka5slb3x93ky1k8s1i7710qdpbwlq1fi-ros-humble-pinocchio-2.6.21-r1.drv»

nix-repl> outputs.legacyPackages.x86_64-linux.humble.python3Packages.pinocchio
«derivation /nix/store/gac53bw2ryvinpyf40rfwc3b5mpkskgw-pinocchio-3.2.0.drv»

**Comment by hacker1024** [comment]
This occurs because ROS packages are in one scope, while Python packages are in another. Using a ROS `callPackage` will shadow Nixpkgs packages from the scope above, but not packages in the Python scope.

i.e.

```nix
rosPackages.rolling.callPackage ({
  buildRosPackage
, pinocchio # Shadowed by the ROS definition
, python3Packages # Not defined by this overlay, so straight from Nixpkgs
}: buildRosPackage { /* ... */ }) { };
```

It would probably be a good idea to check for existing packages either in Superflore or in a step after running it.

---

rtabmap is another shadowed package that would benefit from this.

PR #537, opened 2024-12-08
libfranka: add missing pinocchio dependency

nim65s:
Hi,

libfranka recently added the pinocchio dependency, without declaring it in package.xml, so build fails: https://hydra.iid.ciirc.cvut.cz/build/96125/log/tail

Here is a trivial fix, which build correctly.

PS: ref #536: with a nixpkgs version of eigenpy / hpp-fcl / pinocchio, this would have been easier to build :)

### DIFF
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -58,11 +58,12 @@ in with lib; {
   };
 
   libfranka = rosSuper.libfranka.overrideAttrs ({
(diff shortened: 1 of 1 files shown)


nim65s:
ref. https://github.com/frankaemika/libfranka/issues/170

lopsided98:
It looks like this has been fixed upstream now.

PR #548, opened 2024-12-15
[fix] gazebo: pin ffmpeg to ffmpeg_4

muellerbernd:


lopsided98:
Fixed instead by updating Gazebo

Issue #541, opened 2024-12-13
Some `moveit-ros-*` package builds hang. Relevant to CI issues?


liarokapisv:
Due to the caching issues, I recently had to build many packages by source and I noticed that a few packages stand out. Specifically, packages from `moveit-ros`, like `moveit-ros-move-group` and `moveit-ros-sensors` for some reason take a ton of time to build, I pretty much gave up after 2 hours. It seems these also time-out in the Hydra instance which may be relevant to the CI issues as well. The issue appears at the end of the build phase and is probably linker related, for example see the logs at https://hydra.iid.ciirc.cvut.cz/build/297469/nixlog/1

I put together a minimal flake repo with the proper lockfile to demonstrate this:

```bash
nix build github:liarokapisv/nix-ros-overlay-long-builds-poc#moveit-ros-move-group
```

The build does not complete for me. Same happens if building manually using `nix develop`:

```bash
nix develop github:liarokapisv/nix-ros-overlay-long-builds-poc#moveit-ros-move-group
runPhase unpackPhase
cmake -B build
VERBOSE=1 cmake --build build -j`nproc`
```

Using verbose output for cmake results in the following comand which is what my build gets stuck on:

```bash

store/zznja5f8v3jafffyah1rk46vpfcn38dv-gcc-wrapper-13.3.0/bin/g++
-Wno-deprecated-declarations
-Wno-deprecated-copy
-O3
-DNDEBUG
-Wl,--no-as-needed /nix/store/g9gx7dpfqiy7a1szjsp2jjgihq86lcx8-ros-jazzy-backward-ros-1.0.6-r1/lib/libbackward.so
-Wl,--as-needed CMakeFiles/list_move_group_capabilities.dir/src/list_capabilities.cpp.o
-o list_move_group_capabilities
-Wl,-rpath,/tmp/nix-ros-overlay-long-builds-poc/moveit2-release-release-jazzy-moveit_ros_move_group-2.10.0-1/build:libmoveit_move_group_capabilities_base.so.2.10.0
-lfmt
/nix/store/akdz5s580ych6gq7y7w4c6872srgdx45-ros-jazzy-moveit-ros-planning-2.10.0-r1/lib/libmoveit_constraint_sampler_manager_loader.so.2.10.0
/nix/store/akdz5s580ych6gq7y7w4c6872srgdx45-ros-jazzy-moveit-ros-planning-2.10.0-r1/lib/libmoveit_cpp.so.2.10.0
[...]

PR #550, opened 2024-12-17
remove `--as-needed` propagated flag from backward-ros

liarokapisv:
Fixes https://github.com/lopsided98/nix-ros-overlay/issues/541

Removing the flag relieves the linker in dependent packages like `ros-moveit-*` with many dependencies, since it would otherwise need to search all propagated libraries for symbols.

### DIFF
--- a/distros/ros2-overlay.nix
+++ b/distros/ros2-overlay.nix
@@ -35,6 +35,22 @@ rosSelf: rosSuper: with rosSelf.lib; {
     '';
   });
 
+  backward-ros = rosSuper.backward-ros.overrideAttrs ({ postPatch ? "", ... }: {
+
+    # The `--as-needed` flag directs the linker to search all libraries specified
+    # during its invocation to identify which ones contain the symbols required by the binary.
+    #
+    # Due to the nixpkgs binutils wrapper and the reliance to `propagatedBuildInputs`,
+    # the overlay often propagates an excessive number of libraries.
+    #
+    # As a result, the `--as-needed` flag can significantly increase the time the linker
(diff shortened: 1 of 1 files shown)


wentasah (changes requested):
Can you make the sed expression more understandable as suggested below? Also add a comment explaining why this is needed. A sentence similar to your commit message should be sufficient.

And it seems that this breaks backward-ros in noetic. So this should go into `ros2-overlay.nix`.

wentasah:
Otherwise this looks good. It [fixed 45 packages](https://hydra.iid.ciirc.cvut.cz/eval/1397?compare=develop-all-x86_64&full=0#tabs-now-succeed). 

I was wondering why the `--as-needed` is there at all. It was added in [this commit](https://github.com/pal-robotics/backward_ros/commit/485c6e9ef9bf37e60abc0a967db77708f910ddf2#diff-28ddc9309d8ab162dc9c2ca02fa5ca35cea621012637ccf597f79fb4126c87f6R14), which seems to be the first ROS-related commit in the repo. Maybe, we could try submitting a PR removing that line at upstream. I guess it was some hack needed back in 2017, which may not be relevant today.

liarokapisv:
What is more baffling to me is why this worked in older (post-Summer) nix-ros-overlay builds. The build worked and the `--as-needed` flag was not actually propagated to dependencies. Maybe a cmake version bump changed the behavior?

wentasah (approved):
LGTM. Thanks!

liarokapisv:
Hi there! Is there anything else I can do to help with the merge?

okvik:
Thanks for this, it fixes a bunch of packages that I use.

PR #551, opened 2024-12-18
[fix] tf2-kdl needs pykdl on runtime

muellerbernd:


lopsided98:
https://github.com/ros2/geometry2/pull/745

muellerbernd:
Upstream PR is merged. I think we can close this here.

PR #561, opened 2025-01-16
fix: ros2 build on darwin

stevalkr:
Newer distributions include the `lttng` dependency, which is not available on macOS and Windows platforms. Removing these dependencies automatically disables tracking. Also fixes rviz2 build.

However, ros2 jazzy is still not fully functional on macOS with this commit. Further investigation is required.

### DIFF
--- a/distros/ros2-overlay.nix
+++ b/distros/ros2-overlay.nix
@@ -235,7 +235,7 @@ rosSelf: rosSuper: with rosSelf.lib; {
   }: {
     dontWrapQtApps = false;
     nativeBuildInputs = nativeBuildInputs ++ [ self.qt5.wrapQtAppsHook ];
-    qtWrapperArgs = qtWrapperArgs ++ [
+    qtWrapperArgs = qtWrapperArgs ++ optionals self.stdenv.isLinux [
       # Use X11 by default in RViz2.
       # https://github.com/ros-visualization/rviz/issues/1442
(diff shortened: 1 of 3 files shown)


distros/ros2-overlay.nix
```diff
@@ -248,6 +248,22 @@ rosSelf: rosSuper: with rosSelf.lib; {
...
+      self.darwin.apple_sdk.frameworks.Foundation
+      self.darwin.apple_sdk.frameworks.AppKit
+    ];
+  });
+
+  # lttng-ust and lttng-tools is not available on Darwin, will disable tracking if not available.
+  tracetools = rosSuper.tracetools.overrideAttrs ({
+    propagatedBuildInputs ? [], ...
+  }: {
+    propagatedBuildInputs = if self.stdenv.isDarwin then [] else propagatedBuildInputs;
```
wentasah, line 264: Can you implement this the same way as in `rmw-implementation`? That is:
```nix
    buildInputs = if self.stdenv.isDarwin then
      builtins.filter (p: p.pname != "ros-${p.rosDistro}-rmw-cyclonedds-cpp") buildInputs
    else
      buildInputs;
```
This way, when more packages is later added to propagatedBuildInputs of tracetools, those won't be removed.

wentasah:
I'm closing this. We now have working Darwin build including rviz. See #930.

PR #573, opened 2025-02-12
[fix] jazzy: image-proc

muellerbernd:
see also [upstream PR](https://github.com/ros-perception/image_pipeline/pull/1072)

### DIFF
--- a/distros/jazzy/overrides.nix
+++ b/distros/jazzy/overrides.nix
@@ -180,6 +180,18 @@ in {
     ];
   });
 
+  image-proc = rosSuper.image-proc.overrideAttrs ({
+    patches ? [], ...
+  }: {
+    patches = patches ++ [
+      (self.fetchpatch {
+        url = "https://github.com/ros-perception/image_pipeline/pull/1072.patch";
+        hash = "sha256-eRM+9nj5ZxvNjrY1H4/RqAbcUUeG+g7qUZnQT5wpsJ0=";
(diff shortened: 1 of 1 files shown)


distros/jazzy/overrides.nix
```diff
@@ -180,6 +180,18 @@ in {
     ];
   });
 
+  image-proc = rosSuper.image-proc.overrideAttrs ({
+    patches ? [], ...
+  }: {
+    patches = patches ++ [
+      (self.fetchpatch {
+        url = "https://github.com/ros-perception/image_pipeline/pull/1072.patch";
```
wentasah, line 188: Can you change the url to `https://github.com/ros-perception/image_pipeline/commit/bf7c7c2baddfa958407cca8e57e35caa53deead7.patch`? This is the squashed backported commit which doesn't contain the empty line "noise". And its nix hash will always remain the same, which is not true for PR diffs.

wentasah (changes requested):
Nice, this fixes build failure of 7 packages. Please, change the URL as commented below.

distros/jazzy/overrides.nix
muellerbernd: here you go. thanks for the review.

wentasah (approved):

lopsided98:
I looks like this patch was included in 5.0.8, so this is no longer needed.

PR #581, opened 2025-02-22
Fix the changed hash of the osqp-vendor package

biodranik:
Fixes the following error:

```
error: hash mismatch in fixed-output derivation '/nix/store/11abdpwg3pjbny9bp880hrlpdm7ipan9-osqp.drv':
         specified: sha256-0BbUe1J9qzvyKDBLTz+pAEmR3QpRL+hnxZ2re/3mEvs=
            got:    sha256-DDO7CCDE2mTi/GSaD9wATxnXHvY9Ndz8WI1oQuEbxeI=
```

### DIFF
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -126,7 +126,7 @@ let
         rev = "v0.6.2";
         revVariable = "git_tag";
         fetchgitArgs = {
-          hash = "sha256-0BbUe1J9qzvyKDBLTz+pAEmR3QpRL+hnxZ2re/3mEvs=";
+          hash = "sha256-RYk3zuZrJXPcF27eMhdoZAio4DZ+I+nFaUEg1g/aLNk=";
         };
       })
     ];
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -127,7 +127,6 @@ let
         revVariable = "git_tag";
         fetchgitArgs = {
           hash = "sha256-0BbUe1J9qzvyKDBLTz+pAEmR3QpRL+hnxZ2re/3mEvs=";
-          leaveDotGit = true;
         };
       })
     ];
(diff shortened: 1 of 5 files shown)


biodranik:
@lopsided98 do you have any ideas why [patchExternalProjectGit](https://github.com/lopsided98/nix-ros-overlay/blob/d3e470593e0c9f66e01101d56aa7a0ecef137e95/lib/default.nix#L46) may return a different hash [here](https://github.com/lopsided98/nix-ros-overlay/blob/d3e470593e0c9f66e01101d56aa7a0ecef137e95/distros/distro-overlay.nix#L129) ? It reproduces only on one Ubuntu aarch64 VM.

wentasah:
Confirming that this fixes the problems. Thanks!

Unfortunately, my Hydra doesn't catch these failures, because the packages depend on insecure freeimage and are therefore not evaluated by Hydra. I'm working on a solution to automate updates like these, so that we won't need to do this manually.

wentasah:
The above comment belonged to #585. But this PR also works.

No idea why the hash changes.

lopsided98:
The problem here is probably `leaveDotGit = true`, which tends to make things non-reproducible.

biodranik:
@lopsided98 right, others also mention hash mismatches with leaveDotGit here: https://discourse.nixos.org/t/keep-git-folder-in-when-fetching-a-git-repo/8590/2

Is there a more robust way?

biodranik:
Here is the original issue: https://github.com/NixOS/nixpkgs/issues/8567

Any suggestions for workarounds?

okvik:
@wentasah I don't recall that there was a functional reason I put that in.

Most likely I simply found it useful to have a git repo around during development and didn't know that this could cause issues. Or, I might have needed a git repo for some earlier approach to fixing the build issue.

In any case, I don't see a reason now to keep it in if everything builds fine.

biodranik:
Thanks for the details! I've removed unneded leaveDotGit and fixed hashes. It works now on my VM. Can please someone test that if works on other machines (and maybe also check/fix hashes for humble/rolling)?

PR #584, opened 2025-02-23
Fix example-ros2-gz

ck3d:
Update packages in overrides of jazzy.

Following command can be used to reproduce the issue:
`NIXPKGS_ALLOW_INSECURE=1 nix develop --impure .#example-ros2-gz`


wentasah:
This works well and compared to #585, it also fixes `gz-launch-vendor`.

PR #590, opened 2025-03-01
jazzy: Fix failing patchelf for gz-gui-8

biodranik:
Fixes
```
> patchelf: getting info about '/nix/store/44picdxg6lxgn6d3f3ndipwc9df87a6z-ros-jazzy-gz-gui-vendor-0.0.5-r1/lib64/gz-gui-8/plugins/libGrid3D.so': No such file or directory
```


lopsided98:
Seems fine, but I am unable to reproduce this. Is this on aarch64 or something?

biodranik:
I don't remember exactly, but likely yes. I tried it on both aarch64 and x86_64.

PR #595, opened 2025-03-08
Noetic fixes

wentasah:
This [fixes 143 failed builds](https://hydra.iid.ciirc.cvut.cz/eval/2906?compare=lopsided98-develop&full=0) in noetic.

### DIFF
--- a/distros/noetic/overrides.nix
+++ b/distros/noetic/overrides.nix
@@ -262,6 +262,15 @@ in {
     buildInputs = buildInputs ++ [ self.opencl-clhpp ];
   });
 
+  simple-message = rosSuper.simple-message.overrideAttrs ({
+    postPatch ? "", ...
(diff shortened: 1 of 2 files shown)


wentasah:
More fixes added. Compared to `develop`, this allows building [453 more packages](https://hydra.iid.ciirc.cvut.cz/eval/2913?compare=lopsided98-develop&full=0). The most frequent reason for build failure was Boost 1.87. Some of the packages were fixed by upstream, but many not. For some packages, such as `gazebo_11` (which is already EOL) and quite a few others, the required changes seem non-trivial so I decided that using Boost 1.86 for Noetic might be sufficient, given that Noetic will be EOL in two months.

@lopsided98 If you don't agree, you can merge this except the last two commits. This would fix about 200 packages.

PR #591, opened 2025-03-04
[fix] nav2 packages

muellerbernd:
With this changes `navigation2` compiles for jazzy and humble.

PR #606, opened 2025-03-28
Update Turtlebot3-gazebo example to use Humble distro

mdrodrigo:
We changed the distro from Noetic to Humble for the Turtlebot3-gazebo example. While doing that, we found an error related to boost and Qt5, which we fixed and tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,9 @@ nix-shell \
   --option trusted-public-keys 'cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ros.cachix.org-1:dSyZxI8geDCJrwgvCOHDoAfOm5sV1wCPjBkKL+38Rvo=' \
   '<nix-ros-overlay/examples/turtlebot3-gazebo.nix>'
 # If not on NixOS, nixGL (https://github.com/guibou/nixGL) is needed for OpenGL support
-roslaunch turtlebot3_gazebo turtlebot3_world.launch
+ros2 launch turtlebot3_gazebo turtlebot3_world.launch.py
 # Spawn a new nix-shell in a new terminal and then:
-roslaunch turtlebot3_teleop turtlebot3_teleop_key.launch
+ros2 run turtlebot3_teleop teleop_keyboard
 ```
 
 ### Flakes
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -4,6 +4,15 @@ self:
 rosSelf: rosSuper: let
   inherit (rosSelf) lib;
 in with lib; {
+  # Many packages fail to compile with Boost 1.87
+  boost = rosSelf.boost186;
+
+  # Apply the same override as in distro-overlay.nix
+  boost186 = self.boost186.override {
+    python = rosSelf.python;
+    enablePython = true;
+  };
+
   cyclonedds = rosSuper.cyclonedds.overrideAttrs ({
     patches ? [], ...
   }: {
@@ -29,6 +38,12 @@ in with lib; {
   };
 
   gazebo = self.gazebo_11;
+
+  gazebo-ros = rosSuper.gazebo-ros.overrideAttrs ({
+    nativeBuildInputs ? [], ...
+  }: {
+    nativeBuildInputs = nativeBuildInputs ++ [ self.qt5.qtbase ];
+  });
 
   google-benchmark-vendor = lib.patchExternalProjectGit rosSuper.google-benchmark-vendor {
     url = "https://github.com/google/benchmark.git";
--- a/examples/turtlebot3-gazebo.nix
+++ b/examples/turtlebot3-gazebo.nix
@@ -1,22 +1,19 @@
 # Run:
-# roslaunch turtlebot3_gazebo turtlebot3_world.launch
-# roslaunch turtlebot3_teleop turtlebot3_teleop_key.launch
+# ros2 launch turtlebot3_gazebo turtlebot3_world.launch.py
+# ros2 run turtlebot3_teleop teleop_keyboard
 
 { pkgs ? import ../. {} }:
 with pkgs;
-with rosPackages.noetic;
+with rosPackages.humble;
 with pythonPackages;
 
 mkShell {
   buildInputs = [
     glibcLocales
     (buildEnv { paths = [
-      rosbash
-      turtlebot3-description
+      ros-base
       turtlebot3-teleop
       turtlebot3-gazebo
-      gazebo-plugins
-      xacro
     ]; })
   ];
 


distros/humble/overrides.nix
```diff
@@ -4,6 +4,15 @@ self:
 rosSelf: rosSuper: let
   inherit (rosSelf) lib;
 in with lib; {
+  # Many packages fail to compile with Boost 1.87
+  boost = rosSelf.boost186;
```
wentasah, line 8: I don't think we want to use older boost for Humble. Sooner or later, even boost186 will disappear from nixpkgs and we'd have to update anyway. So it's better to update now than waiting and then solving more accumulated problems.

Which failing packages do you need? If nav2-*, then it might be solved similarly as in #591 for Jazzy.

distros/humble/overrides.nix
```diff
@@ -30,6 +39,12 @@ in with lib; {
 
   gazebo = self.gazebo_11;
 
+  gazebo-ros = rosSuper.gazebo-ros.overrideAttrs ({
+    nativeBuildInputs ? [], ...
+  }: {
+    nativeBuildInputs = nativeBuildInputs ++ [ self.qt5.qtbase ];
```
wentasah, line 45: Shouldn't this be just `buildInputs`? It makes the difference in cross compilation.

wentasah (changes requested):
Thanks for the PR.

If we change the example to ROS 2 now, we won't have an easy way to test whether ROS 1 still works. So I'd prefer to create a new example.

Regarding the other change, see the comments in the code.

distros/humble/overrides.nix
otavio: I think you are right. We will look at this on Monday.

otavio:
Ok, I think that it makes sense to create a new example and preserve the ROS 1 example for testing purposes.

distros/humble/overrides.nix
mdrodrigo: We changed it and it worked. Thanks

distros/humble/overrides.nix
mdrodrigo: We made the change to use the `-Wno-error=array-bounds` flag in `nav2-costmap-2d`, and it works fine.
However, we encountered another error related to the `boost` library and `ros-humble-gazebo-plugins`. Can you help us fix this issue?

```
ros-humble-gazebo-plugins> [ 17%] Building CXX object CMakeFiles/gazebo_ros_joint_state_publisher.dir/src/gazebo_ros_joint_state_publisher.cpp.o
ros-humble-gazebo-plugins> [ 17%] Building CXX object CMakeFiles/multi_camera_plugin.dir/src/multi_camera_plugin.cpp.o
ros-humble-gazebo-plugins> In file included from /nix/store/jm5d6cvg9imwpwzb6an6rq5gf6ryp3r8-gazebo-11.15.1/include/gazebo-11/gazebo/transport/Connection.hh:21,
ros-humble-gazebo-plugins>                  from /nix/store/jm5d6cvg9imwpwzb6an6rq5gf6ryp3r8-gazebo-11.15.1/include/gazebo-11/gazebo/transport/transport.hh:3,
ros-humble-gazebo-plugins>                  from /build/gazebo_ros_pkgs-release-release-humble-gazebo_plugins-3.9.0-1/src/gazebo_ros_ray_sensor.cpp:17:
ros-humble-gazebo-plugins> /nix/store/vbnp9cdf34ahixgr5ni4j2s9fhlcl7wx-tbb-2020.3-dev/include/tbb/task.h:21:139: note: '#pragma message: TBB Warning: tbb/task.h is deprecated. For details, please see Deprecated Features appendix in the TBB reference manual.'
ros-humble-gazebo-plugins>    21 | #pragma message("TBB Warning: tbb/task.h is deprecated. For details, please see Deprecated Features appendix in the TBB reference manual.")
ros-humble-gazebo-plugins>       |                                                                                                                                           ^
ros-humble-gazebo-plugins> /nix/store/jm5d6cvg9imwpwzb6an6rq5gf6ryp3r8-gazebo-11.15.1/include/gazebo-11/gazebo/transport/Connection.hh:412: error: 'boost::asio::ip::resolver::iterator' has not been declared
ros-humble-gazebo-plugins>   412 |                   boost::asio::ip::tcp::resolver::iterator _endPointIter);
ros-humble-gazebo-plugins> In file included from /nix/store/jm5d6cvg9imwpwzb6an6rq5gf6ryp3r8-gazebo-11.15.1/include/gazebo-11/gazebo/transport/transport.hh:5:
ros-humble-gazebo-plugins> /nix/store/jm5d6cvg9imwpwzb6an6rq5gf6ryp3r8-gazebo-11.15.1/include/gazebo-11/gazebo/transport/IOManager.hh:45: error: 'io_service' in namespace 'boost::asio' does not name a type; did you mean 'use_service'?
ros-humble-gazebo-plugins>    45 |       public: boost::asio::io_service &GetIO();
```

Let know us if you need more information.
Regards

mdrodrigo:
Hi @wentasah

We have implemented the proposed changes by creating a new example for the Humble ROS distribution while keeping the original example for testing with the Noetic ROS distribution.

Please let us know if you would like any additional changes to help the project.

Regards.

cc: @otavio

distros/humble/overrides.nix
wentasah: As far as I understand, gazebo-11 is after  its [end of life](https://github.com/gazebosim/gazebo-classic). So this example should be ported to newer `gz-*` packages. We already have an [example with that](https://github.com/lopsided98/nix-ros-overlay/blob/develop/examples/ros2-gz.nix), so you can get inspiration from there.

README.md
```diff
@@ -17,27 +17,46 @@ Want to use ROS, but don't want to run Ubuntu? This project uses the power of [N
...
 
-Turtlebot 3 simulation in Gazebo:
+Turtlebot 3 simulation in Gazebo on noetic distro:
 ```
 nix-shell \
   -I nix-ros-overlay=https://github.com/lopsided98/nix-ros-overlay/archive/master.tar.gz \
   --option extra-substituters 'https://ros.cachix.org' \
   --option trusted-public-keys 'cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ros.cachix.org-1:dSyZxI8geDCJrwgvCOHDoAfOm5sV1wCPjBkKL+38Rvo=' \
-  '<nix-ros-overlay/examples/turtlebot3-gazebo.nix>'
+  '<nix-ros-overlay/examples/noetic-turtlebot3-gazebo.nix>'
```
wentasah, line 26: Please don't change the names of existing files (and attributes below). Users may rely on these names in their Nix expressions and this would break them.

examples/humble-turtlebot3-gazebo.nix
```diff
@@ -0,0 +1,23 @@
...
+with rosPackages.humble;
+with pythonPackages;
+
+mkShell {
+  buildInputs = [
+    glibcLocales
+    (buildEnv { paths = [
+      ros-base
+      turtlebot3-teleop
+      turtlebot3-gazebo
```
wentasah, line 16: This package depends on gazebo_ros_pkgs, which is [deprecated](https://index.ros.org/p/gazebo_ros_pkgs/#humble). They recommend to follow https://gazebosim.org/docs/latest/gazebo_classic_migration. Seems like https://gazebosim.org/docs/latest/migrating_gazebo_classic_ros2_packages/ might be what you need.

I also wonder whether #591 contains what you intend to do here or not. There is a new example [ros2-turtlebot4-gz.nix](https://github.com/lopsided98/nix-ros-overlay/pull/591/files#diff-c76790b502439e005313c9d13f55084aa07afec03b708313950637f334fadf91).

wentasah (changes requested):

distros/humble/overrides.nix
otavio: Absolutely, but as we know, the TurtleBot3 does not support the new Gazebo, so for this to work we need to use the classic version. The most important thing for us is to try to understand how to solve this kind of problem.

We created a testing TurtleBot4 example which we can send as well, but as a learning exercise we would like to understand how we would identify which package we would need to fix to work around that boost issue. The only solution that we could come up with was using the old version of boost which is completely understandable to not be advisable.

wentasah: This is a complicated situation. Normally, one would create a patch for gazebo 11 fixing the boost issues and send it upstream. But since upstream is EOL, this wouldn't work. The other option is, as you write, compile everything against older boost, but this might create problems with other packages in the future. Third option would be to carry the gazebo patch in this repo, but I think that @lopsided98 doesn't want to maintain here packages/patches, which are not supported upstream.

I see that the [TurtleBot3 overview page](https://emanual.robotis.com/docs/en/platform/turtlebot3/overview/) contains:
> As a priority, TurtleBot3 will receive full support for ROS 2 Humble, with comprehensive example implementations set for release in Q1 2025. In Q2, support will expand to ROS 2 Jazzy and Gazebo Sim, ensuring seamless integration with the latest advancements in the ROS ecosystem and simulation environments.

So maybe, if you wait a bit, the support for new gazebo will be developed (maybe, you can help them with that?). In the meantime, I'd suggest to use either older version of this overlay or your local fork with downgraded Boost.

otavio: Yes, I noted that in the project documentation, considering this point that they are going to fix this in near future, I propose a different solution that would be to do the downgrade in the Boost library for now, and when they update the TurtleBot3 example to work with the new Gazebo, then we remove that change. Would you accept this?

My point here is that it would be good to have more packages working and being tested, so if there is no current solution for now and the preference for not carrying fixes by yourself, which I fully understand, I think that would be a good solution at this moment.

README.md
otavio: I was discussing that with @mdrodrigo and he brought a valid point that the example was not working as nobody noticed that it was failing to build. So I'm pretty sure that users are not relying on this otherwise we would have received issues about it failing to build and run. Considering that I think that would be better to rename the attribute so it is clear which version it is using and causes less confusion for users when testing it.

What do you think?

examples/humble-turtlebot3-gazebo.nix
otavio: The proposed pull request certainly tackles some of the issues, and I appreciate that effort! However, I believe it's really important for the layer to create a welcoming experience for new users. If there's a common issue with building widely used packages, it might not provide the best first-time experience for them. I encourage you to take a closer look at my earlier comments and consider the options I've suggested- they're intended to help us all move forward together!

PR #612, opened 2025-04-03
release.nix: Fix eval error when evaluating without arguments

wentasah:
Without this, when running `nix-eval-jobs release.nix`, the following error occurs early, preventing generating useful output:

    error:
       … while evaluating recurseForDerivations

       … while calling the 'removeAttrs' builtin
         at /home/wsh/src/ros/nix-ros-overlay/release.nix:18:26:
           17|   inherit (builtins) mapAttrs attrNames filter listToAttrs readDir;
           18|   cleanupDistro = (_: a: removeAttrs a [
             |                          ^
           19|     "lib"

       … while evaluating the first argument passed to builtins.removeAttrs

       error: expected a set but found a Boolean: true

With this change, `nix-eval-jobs release.nix` tries to evaluate all packages.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -14,8 +14,8 @@ in
 let
   pkgs = import nix-ros-overlay { inherit nixpkgs system; };
   inherit (pkgs.lib) isDerivation filterAttrs;
-  inherit (builtins) mapAttrs attrNames filter listToAttrs readDir;
-  cleanupDistro = (_: a: removeAttrs a [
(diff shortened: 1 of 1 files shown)


wentasah:
I no longer use `release.nix` here. My hydra uses release.nix from [here](https://github.com/wentasah/nix-ros-hydra). But let's merge this to fix the eval error, if someone needs it.

PR #619, opened 2025-04-15
Fix/openvdb vendor

mbeutelspacher:
Apply patchAmendVendorGit to openvdb-vendor.

For that I needed to ensure that patchAmendVendorGit removes existing ``VCS_TYPE`` from the ``CMakeLists.txt``

### DIFF
--- a/distros/humble/openvdb-vendor/vendored-source.json
+++ b/distros/humble/openvdb-vendor/vendored-source.json
@@ -1,0 +1,5 @@
+{
+  "url": "https://github.com/AcademySoftwareFoundation/openvdb.git",
+  "rev": "v10.0.1",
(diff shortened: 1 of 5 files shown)


wentasah:
Hi, this looks good.

I'm currently extending `patchAmentVendorGit` to work well for `zenoh-cpp-vendor` and this would conflict with this PR. So I took your VCS_TYPE fix and added it to #620. I'm not sure which PR gets merged first. If #620, `vendored-source.json` here would need to be regenerated.

wentasah:
I think we can merge this. The package will have to be built from source either way.

It seems the only ROS package that depends on `openvdb-vendor` is `spatio-temporal-voxel-layer`, but currently it cannot find openvdb - neither from `openvdb-vendor` nor from nixpkgs. So I don't think one option is better than the other. If people have problems with the vendored version, we can always switch to the nixpkgs one later.

PR #558, opened 2025-01-10
[jazzy] zenoh-cpp-vendor WIP

muellerbernd:
I have tested the somewhat new middleware [rmw_zenoh](https://github.com/ros2/rmw_zenoh) in a local workspace. Now I try to get the corresponding nix packages running. 
When try to build them:
```bash
nix build .\#jazzy.zenoh-cpp-vendor --accept-flake-config
```
Colcon gets invoked, then colcon executes cargo and cargo cannot download stuff because it is sandboxed. According to the [nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#compiling-non-rust-packages-that-include-rust-code) we would have to set the `cargoDeps`. 
I saw that in #501 was also done some work to build rust nodes.

PR #620, opened 2025-04-16
Allow generating vendored-source.json for packages with multiple ament_vendor calls

wentasah:
This is needed at least for `zenoh-cpp-vendor`, which is being prepared in #558.

The format of `vendored-source.json` is changed from a single object to an array of objects. The order of objects corresponds to the order of `ament_vendor` calls in the patched `CMakeLists.txt`.

`ament_cmake_vendor_packageConfig.cmake` was updated to generate such a format automatically and `patchAmentVendorGit` was updated to understand that format and patch all call sites in the correct order.

The `vendored-source.json` files are regenerated automatically and there are no changes except addition of extra brackets.


wentasah:
Fixed patching to build `zenoh-cpp-vendor` correctly.

wentasah:
Even this is not good. This version fails with `rviz-ogre-vendor`, which contains several calls to `ament_vendor`, guarded with different `if` conditions. This version patches only the first occurrence, which is for WIN32. So we'll have to update the patching code to sometimes patch all occurrences (the old behavior) and sometimes one by one (this PR up to now).

wentasah:
Now it works for all packages in the overlay as well as for `zenoh-cpp-vendor`. See https://hydra.iid.ciirc.cvut.cz/eval/3871?compare=lopsided98-develop&full=0.

muellerbernd:
> Now it works for all packages in the overlay as well as for `zenoh-cpp-vendor`. See https://hydra.iid.ciirc.cvut.cz/eval/3871?compare=lopsided98-develop&full=0.

Thanks for work. Now everything works fine. As soon as this gets merged I will update #558 .

lib/default.nix
```diff
@@ -90,33 +90,57 @@
...
-      echo >&2 ${lib.escapeShellArg errMsg}
-      exit 1
-    '') + postPatch;
+    postPatch = (
+      if pathExists vendoredSourceJson then
+        # Remove all existing VCS_TYPE lines
+        ''
+          sed -i '/VCS_TYPE \(git\|zip\|svn\|path\)/d' ${lib.escapeShellArg file}
+        '' + (
+          if length sourceInfos == 1 then
```
lopsided98, line 117: This seems a bit brittle. Not that big a deal since we only have to handle a limited number of examples, but as an alternative, would it be feasible to match against the original `VCS_URL` value when patching?

wentasah: Unfortunately not. CMakeFile authors are very creative and use variables in URLs like:

    VCS_URL https://github.com/gazebosim/${GITHUB_NAME}.git

I was also looking at whether `cmake` can tell us the line number from where a macro (of a function) was called so that we can match based on that, but nothing like this seems to exist.

lopsided98: I guess it could be done if there were some way to wrap the `ament_vendor()` call with our own macro that replaces the `VCS_URL` parameter.

wentasah: I think the problem to solve here would be how to call the original `ament_vendor` implementation from the wrapper with the same name. I don't see any means for macro namespacing in cmake. Maybe, we can use Nix to patch the original implementation to have a different name and call that from the wrapper.

wentasah:
I'm reworking this based on above discussion with @lopsided98, but it takes me longer than expected due to other duties. I'd like to have it ready this week. I'll then look at zenoh-cpp-vendor because without it, we won't beady for Kilted release.

muellerbernd:
@wentasah thanks for your work. I have updated my [test branch](https://github.com/muellerbernd/nix-ros-overlay/tree/pr-620-additions) to be up to date with this PR. for humble and jazzy `zenoh-cpp-vendor` do compile.

wentasah:
@lopsided98 This is an updated version, which wraps calls to `ament_vendor` and should be more reliable than previously used patching of `CMakeLists.txt` files with `sed`. For details see individual commit messages and comments in the code. It doesn't seem to cause any [breakage compared to the develop branch](https://hydra.iid.ciirc.cvut.cz/eval/4356?compare=lopsided98-develop&full=0).

@muellerbernd Thanks for quick testing.

PR #634, opened 2025-05-27
humble: ros2-ouster: add missing build inputs

aboseley:


wentasah (changes requested):
Hi and thanks for the patch. It seems that the libraries don't need to be in `propagatedBuildInputs`. `buildInputs` is probably sufficient, because the header files included in `ros2-ouster` output do not include headers from the mentioned libraries directly.

wentasah:
Closing in favor of #676.

PR #637, opened 2025-06-12
humble: fix plotjuggler build

nim65s:
Hi,

It seems that Davide is having some fun with CMake :)

I tried to fix v3.10.1 without success. v3.10.6 is better.
With this I can start `./result/lib/plotjuggler/plotjuggler`

### DIFF
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -242,7 +242,7 @@ in with lib; {
 
 
   plotjuggler-ros = rosSuper.plotjuggler-ros.overrideAttrs ({
-    patches ? [], nativeBuildInputs ? [], ...
+    patches ? [], postPatch ? "", nativeBuildInputs ? [], ...
   }: {
     patches = patches ++ [
       # Fix detection of ROS 2 Humble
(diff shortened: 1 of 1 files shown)


wentasah:
Probably also OK, but it seems that 3.10.7 is out and contain similar (if not the same) fixes. It seems we have some issues with automatic updates of the overlay. I'll look at that first and it may make this PR unnecessary.

wentasah:
> It seems that Davide is having some fun with CMake :)

It really must have been fun. I played with this a bit. 3.10.6 tag that you use no longer exists in the repository. Fortunately, 3.10.6-r1 is already released in humble so your patch works after removing the `src` override. But 3.10.7 is out and is propagated to Jazzy but not yet to Humble. So in Jazzy, I had to remove the `if(NOT TARGET mcap)` patch, which is fixed there. And after 3.10.7 it changed further (https://github.com/facontidavide/PlotJuggler/commit/c3e443a0f7926d30b16a79017f1d83df9f919685), which will probably make `lz4` dependency not needed.

So this will need to be updated. I'd also add the same fix to Jazzy. Kilted is still on 3.10.5 but it will need the same after the update. Currently, we're lagging behind ROS a bit so hopefully, it will not be a problem to wait a bit more until all of this settles down.

I also wonder whether the `find_or_download_data_tamer` thing should be submitted upstream. What do you think?

nim65s:
I guess it would be wiser to let this stabilize a bit.
But if someone need a working version sooner I am fine fighting this beast again: it is a good challenge :)

Otherwise we could propose some help upstream, but I suspect they are just directly testing stuff on the buildfarm.

nim65s:
3.10.11 is in #654

nim65s:
And it builds 🎉

Issue #633, opened 2025-05-19
ROS2 rtabmap-ros


muellerbernd:
I am currently working with RTABMap and the ROS version did not build. I have created a new [branch](https://github.com/muellerbernd/nix-ros-overlay/tree/fix-rtabmap) in my fork of `nix-ros-overlay`.

With these changes `rtabmap-ros` compiles but for that I had to make some changes:
- use an updated version of `nixpkgs`
- compile vtk with `-DVTK_MODULE_ENABLE_VTK_GUISupportQt` because the rtabmap-ros packages needed that
  ```nix
  vtk = super.vtk.overrideAttrs ({
    cmakeFlags ? [], nativeBuildInputs ? [], ...
  }: {
    cmakeFlags = cmakeFlags ++ ["-DVTK_MODULE_ENABLE_VTK_GUISupportQt:STRING=YES"];
    nativeBuildInputs = nativeBuildInputs ++ [ self.qt5.wrapQtAppsHook self.qt5.full ];
  });
  ```
I took inspirations from the discussion in #536.

muellerbernd:
but with these changes `ouster-ros` does not compile because of the newer `eigen` version.

PR #639, opened 2025-06-25
Fix rtabmap

muellerbernd:
building on top of #633


wentasah:
I've merged the `libpointmatcher` fix. What's the status of the other commits? Can you rebase them to develop and squash related WiP commits to a single final one?

Also, if something needs newer nixpkgs, do it in a new PR. Don't update nixpkgs in the PR and just say that the PR needs new nixpkgs. Updating nixpkgs must be done carefully, since it usually breaks many packages. After I update nixpkgs, I'll merge these PRs.

PR #640, opened 2025-06-25
workflows: Truncate PR message if it's longer than GitHub limit

wentasah:
Recently, updates started failing (e.g. https://github.com/lopsided98/nix-ros-overlay/actions/runs/15634988130/job/44047768366). This is either because GitHub decreased the limit for PR message length or ROS became bigger than ever before (perhaps due to the Kilted release) or both.

This PR blindly truncates the message if it is too long. Hopefully, it will not happen too often. If it happens, we can recognize it from GitHub Action log message.

This change was tested in my fork. The truncated PR generated by the proposed Action is https://github.com/wentasah/nix-ros-overlay/pull/23.


Amronos:
> It says: `maximum is 65536 characters`.

That is somewhat misleading, please have a look at the links I sent! Thanks!

Issue #544, opened 2024-12-13
Adding all dependencies to `propagatedBuildInputs` is not sustainable.


liarokapisv:
The heavy usage of `propagatedBuildInputs` means that projects with many dependencies end up collecting a ton of gcc flags through the `cc-wrapper`.  

This includes - for each dependency in the dependency chain - the include folder path and the lib folder path together.
Each path is ~80 characters on average.

When using a dev-shell for a workspace, it will pickup a ton of transitive dependencies from all the workspace packages.

As an example, my devshell environment ends up with ~260 dependencies (because they are transitive).

For some reason these get a x3 duplication in the cc-wrapper  invocation, probably due to a cc-wrapper setup-hook.sh omition when `propagatedBuildInputs` or something, I am still investigating the reason.

This means that any gcc invocation ends up with x3 x260 x2 x80 = 124800 kb worth of flags. 

Add all the extra `-system` and additional gcc crafts and we end up hitting https://github.com/NixOS/nixpkgs/issues/41340

This was not an issue for me with an older `master` commit, I am guessing that some packages just added a few extra transitive dependencies and it managed to tip the balance.

liarokapisv:
It appears that the amplification happens due to `cc-wrapper` setup hooks being injected at different levels. Specifically the problem seems to appear due to `clang` being used for the various ament tools.

See relevant issue at https://github.com/NixOS/nixpkgs/issues/364984.

hacker1024:
To my understanding, we end up with many `propagatedBuildInputs` because it's often impossible to know if a dependency actually needs to be propagated or not. [REP 149](https://www.ros.org/reps/rep-0149.html#dependency-tags) actually defines a wide range of dependency types that can map pretty nicely to Nixpkgs's, but a significant amount of packages use the basic [`depend`](https://www.ros.org/reps/rep-0149.html#depend-multiple) tag, which is not helpful.

(See [here](https://github.com/lopsided98/superflore/blob/nixos-support/superflore/generators/nix/nix_package.py#L85-L111) for the current Nix generation logic.)

I wonder how necessary this actually is. What would happen if we tried to use `buildInputs` instead? If there are certain common dependencies that do need to be propagated, could we get away with maintaining a list of them and only propagating packages from that whitelist by default? At the very least, we can use a blacklist - I doubt `rclcpp` ever needs to be propagated, for example, but it often is.

[...]

PR #642, opened 2025-06-26
Allow updating the overlay manually

wentasah:
Given that the update GitHub Action is failing for multiple weeks (fix is in #640), it might be useful for users to perform the update locally. With this PR, one can update the overlay locally by running:

    nix run .#update-overlay

This executes more or less the same commands as the GitHub Action performing periodic updates. The second commit updates the action to use the same instead of duplicating the code in `update.yaml`. This is being tested in https://github.com/wentasah/nix-ros-overlay/actions/runs/15906983859/job/44864392480.

This functionality might also be useful for solving issues like #544.

### DIFF
--- a/.github/workflows/update.yaml
+++ b/.github/workflows/update.yaml
@@ -18,6 +18,15 @@ jobs:
       - name: Install Superflore
         run: |
           nix build .#update-overlay
+      - name: Restore ROS tarball cache
+        uses: actions/cache/restore@v4
+        with:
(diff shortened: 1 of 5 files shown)


wentasah:
Fixed PR creation. Testing it again in https://github.com/wentasah/nix-ros-overlay/actions/runs/15911788222/job/44880710886. If this fails just due to insufficient permissions, it's correct. It doesn't have permission to create PRs in this repo.

PR #657, opened 2025-09-09
update jazzy mcap-vendor override

nim65s:
Hi,

This fix plotjuggler build on jazzy.

mcap-vendor otherwise seems OK on humble and rolling.

I guess those `patchVendorUrl` are manually updated, otherwise feel free to close this PR.

Also, if they are manual, it could be worth working on some automation.

### DIFF
--- a/distros/jazzy/overrides.nix
+++ b/distros/jazzy/overrides.nix
@@ -142,8 +142,8 @@ in {
   };
 
   mcap-vendor = lib.patchVendorUrl rosSuper.mcap-vendor {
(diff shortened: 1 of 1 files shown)


wentasah:
Regarding automated updates. We can do it if packages use `ament_vendor` command in their `CMakeLists.txt`. If not, CMakeLists.txt authors are usually very creative and everybody implements source vendoring differently. This means you would have to create different automation for every package.

What would make sense is, if the package already uses ament, convert their CMakeLists to use `ament_vendor`. I think that in case of `mcap_vendor` it should be possible.

PR #658, opened 2025-09-09
jazzy: fix usb-cam

znaniye:
Since this change was made after the last release, the package was broken. I applied the fix patch so it builds correctly.


wentasah:
Thank you for the fix. Next time, could you also add a short comment explaining why the patch is needed? For this PR, I added it myself.

PR #666, opened 2025-09-13
zenoh_cpp_vendor

wentasah:
This is a set of commits, which finally fixes build of `zenoh_cpp_vendor`. This is important, because many `kilted` packages depend on it.

This took inspiration from #558, cleaned up few things  and added changes required for the current Zenoh version.

### DIFF
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -311,31 +311,54 @@ in with lib; {
     ];
   });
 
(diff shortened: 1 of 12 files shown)


wentasah:
Hmm, I see. But this would probably require https://github.com/NixOS/nixpkgs/commit/112d13c5c7c83593972e786cfa9f6fcef06bdfd9, which is not in nixpkgs we use now. Moving to current nixpkgs-unstable breaks quite a lot of things and I want to work on that in few weeks after I cleanup the repo a bit (getting rid of noetic, etc.).

PR #703, opened 2025-10-15
fix: colcon-core build

znaniye:
Mismatch between `setuptools` versions was causing the `colcon-core` build to fail:  
```shell
Successfully built colcon_core-0.20.1-py3-none-any.whl
Finished creating a wheel...
Finished executing pypaBuildPhase
Running phase: pythonRuntimeDepsCheckHook
@nix { "action": "setPhase", "phase": "pythonRuntimeDepsCheckHook" }
Executing pythonRuntimeDepsCheck
Checking runtime dependencies for colcon_core-0.20.1-py3-none-any.whl
  - setuptools<80,>=30.3.0 not satisfied by version 80.9.0.post0
```

### DIFF
--- a/pkgs/colcon/core.nix
+++ b/pkgs/colcon/core.nix
@@ -39,6 +39,8 @@ let
     pyproject = true;
     build-system = [ setuptools ];
 
+    pythonRelaxDeps = true;
+
     propagatedBuildInputs = [
       distlib
       empy_3


wentasah:
I'm aware of the colcon build failure, but I'm not sure such a solution is what we want. Setuptools incompatibility was declared by colcon [for a reason](https://github.com/colcon/colcon-core/pull/699) and if you ignore it, you'll end up with `colcon build --symlink-install` being broken (I've just tested it).

Solving this should probably be done upstream. I plan to look at this soon, but if you or somebody else want to help, that would be nice.

PR #705, opened 2025-10-16
[fix] mimick-vendor for all distros

muellerbernd:
needed if you want to compile navigation2 from source in your workspace


wentasah:
Thanks. I added a comment that `NIX_CFLAGS_COMPILE` could be removed once the last commit appears in ROS.

PR #718, opened 2025-10-20
[fix] gtsam for rolling and kilted

muellerbernd:


wentasah:
Thanks. Can you add a comment that boost is an optional dependency, but enabled by default? And add a link to https://github.com/borglab/gtsam/pull/2171.

And add it just for kilted and rolling, not for everything in distro-overlay. From the PR it seems that boost may not be needed in the future and if we have it in distro-overlay, it would probably stay there forever.

wentasah:
Thanks. I just squashed the two changes and separated whitespace cleanup to a separate commit.

PR #746, opened 2025-11-21
Gazebo without insecure FreeImage

wentasah:
We have a problem with Gazebo that it needs insecure FreeImage library. Since FreeImage was removed from nixpkgs, [I tried to replace it with stb](https://github.com/gazebosim/gz-common/pull/725), as suggested by upstream. With that change, I'm able to build Gazebo by `colcon` and run it. Also all tests that passed before the change pass after it too. This PR is an unfinished attempt to bring this change to this overlay. `rolling.gz-launch-vendor` builds but more setup is needed to run it and it's too late for me to continue now :-).

@nim65s You definitely have more experience with Gazebo than me and maybe can help here (or in the upstream PR). I also [started preparing the version of the patch for Jazzy](https://github.com/wentasah/gz-common/tree/replace-freeimage-gz5) (another attempt [here](https://github.com/wentasah/gz-common/tree/replace-freeimage-gz5.7.1)), but this is not tested at all yet. My plan is that if it works, it will allow merging #738.

### DIFF
--- a/distros/kilted/overrides.nix
+++ b/distros/kilted/overrides.nix
@@ -97,11 +97,15 @@ in {
 
   gz-cmake-vendor = lib.patchGzAmentVendorGit rosSuper.gz-cmake-vendor { };
 
-  gz-common-vendor = (lib.patchGzAmentVendorGit rosSuper.gz-common-vendor { }).overrideAttrs ({
-    nativeBuildInputs ? [], ...
-  }: {
-    # https://github.com/gazebo-release/gz_common_vendor/pull/2
-    nativeBuildInputs = nativeBuildInputs ++ [ self.pkg-config ];
+  gz-common-vendor = (lib.patchGzAmentVendorGit rosSuper.gz-common-vendor {
+    patchesFor.gz_common_vendor = [
+      (self.fetchpatch2 {
+        # Replace FreeImage dependency with stb (#725 updated for Kilted)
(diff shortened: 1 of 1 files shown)


nim65s:
Maybe they managed to remove it, that would be woderful :)

nim65s:
Ok, so I can now pretty confidently say that removing freeimage from gazebo with your patch works 🎉 

But I could not remove freeimage from ogre-next and have something that work :(
ref. https://github.com/OGRECave/ogre-next/issues/437 

I gave FreeImageRe a couple hours, but, even if it does claim to fix some CVE, there is still a lot of those. And from the packaging side and dependency management point of view, it is in a very bad shape.

wentasah:
I'm not sure, but I think I compiled gazebo against ogre-1.9 (which we have in nix-ros-overlay) and it was possible to compile it without freeimage, by simply overriding with `freeimage = null`. Do you know what are the differences in gazebo when compiled against ogre-1.9 and ogre-next? Performance? Functionality?

wentasah:
Now, the version from Jazzy seems to work too:
```
nix develop github:wentasah/nix-ros-overlay/rolling-gz-no-freeimage#example-ros2-gz-jazzy --option extra-substituters https://attic.iid.ciirc.cvut.cz/ros --option extra-trusted-public-keys ros:JR95vUYsShSqfA1VTYoFt1Nz6uXasm5QrcOsGry9f6Q=
```
Can somebody test it?

PR #749, opened 2025-11-23
CI and caching improvements

wentasah:
CI runs too long and the binary cache seems to be too small to store all what's needed. These commits try to improve the situation and provide more insight into how is the cache used. See the individual commit messages for more information.

### DIFF
--- a/.github/actions/nix-ros-build-action/action.yml
+++ b/.github/actions/nix-ros-build-action/action.yml
@@ -25,6 +25,9 @@ inputs:
     description: Name of the Cachix cache to use
     required: true
     default: ros
+  pin-name:
+    description: Name of the Cachix pin to pin the results
+    required: false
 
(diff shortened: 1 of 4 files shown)


nim65s:
> This means that when some source disappears or gets renamed (both
> happened in the past) our users will not be able to rebuild all
> packages from source.

This can probably be handled by https://www.softwareheritage.org/ as in https://github.com/nix-community/nixpkgs-swh

wentasah:
Thanks for the info. If I understand correctly, we could have a CI job producing `sources.json` and submitting to SW heritage. Do nixpkgs fetchers fallback to fetching sources from SW heritage automatically?

nim65s:
I know that guix fetchers do fallback to SWH.
For nix I think I remember this was not done yet, but I'm not sure.

PR #785, opened 2026-01-30
inherit crocoddyl from nixpkgs

nim65s:
ref. #536


wentasah:
The version of crocoddyl in our locked nixpkgs is 3.0.1 whereas rosdistro (nixpkgs unstable) has 3.2.0. Do you think it will not cause problems to potential users?

Also, ROS has crocoddyl only in rolling, but this PR adds it to all distros. Do you plan to add it there also in ROS?

PR #821, opened 2026-03-18
fix: spacenav for all distros

muellerbernd:


muellerbernd:
can be closed changes from [here](https://github.com/ros/rosdistro/pull/50303) have arrived in superflore.

PR #825, opened 2026-03-23
Adding ros2 numpy

wkaisertexas:
[Ros2 Numpy](https://github.com/Box-Robotics/ros2_numpy) is a helpful library that we use at Cavalier autonomous and always have to clone into an external directory because it does not have an apt package.

This PR adds it to the `nix-ros-overlay` so it can be installed using nix.

### DIFF
--- a/distros/jazzy/ros2-numpy/default.nix
+++ b/distros/jazzy/ros2-numpy/default.nix
@@ -1,0 +1,25 @@
+# Copyright 2026 Open Source Robotics Foundation
+# Distributed under the terms of the BSD license
+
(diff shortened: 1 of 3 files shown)


nim65s:
Hi,

It is not clear to me if adding ROS packages that are not in rosdistro is in scope for this repo or not.

wentasah:
Yes, definitely not this repo. But since there are many non-official ROS packages, I think it would be beneficial to have a repository for them. Something like NUR repo with shared maintenance model and basic CI. I was thinking about starting it some day, but there are still more other things to do :-)

nim65s:
I have almost everything ready for that, I'll try to finish and announce it next week

PR #862, opened 2026-05-31
Allow using buildEnv as ROS underlay

wentasah:
When one has a package X in a Nix development environment (or nix-shell) and a package of the same name appears in a local ROS workspace, previous buildEnv implementation ignored the local package. With this change, the user can set `underlay = true` to make the local package override the package from Nix. In that case, if the ROS workspace has no package named X, the one from Nix will be visible.

Originally I thought, it should be possible to achieve the same without modifying nix-ros-overlay, by just disabling wrapping of binaries (e.g. ros2) in the buildEnv with `wrapPrograms = false`. But for some reason, it doesn't work. It makes some packages not visible at all, which looks like a bug. Until this bug is fixed, we just do what's in this commit.

Moreover, it's not sufficient to change only AMENT_PREFIX_PATH. It seems that `ros2 launch` searches the packages based on PYTHONPATH rather than based on AMENT_PREFIX_PATH. Therefore, changes performed in Python code in a local workspace launched via `ros2 launch` were not visible. We fix that by setting also PYTHONPATH.

And while we're at it, we also change the remaining variables. It seems to be the right thing for common use cases.

---

P.S. I was replacing `--prefix` with `--suffix` in all of my ROS projects in the past. Since I'm lazy to copy that patch to new projects, I made the change configurable and post it here. I add `underlay = true` to the example template, because this is likely the behavior most people expect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Using the overlay in your `flake.nix`-based project could look like this:
             pkgs.colcon
             # ... other non-ROS packages
             (with pkgs.rosPackages.humble; buildEnv {
+              underlay = true;
               paths = [
                 ros-core
                 # ... other ROS packages
--- a/distros/build-env/default.nix
+++ b/distros/build-env/default.nix
@@ -10,11 +10,18 @@
 #
 # By default, all binaries in the environment are wrapped, setting the relevant
 # ROS environment variables, allowing use outside of nix-shell.
+# Wrapping prefixes the existing variables with the buildEnv output, which
+# ensures reproducibility and prevents an incompatible user environment from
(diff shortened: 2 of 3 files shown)


nim65s:
fwiw, I have another solution to this issue, based on `mkShell.inputsFrom`, along with a lib that help generate `mkShell` / `buildEnv`.

It is still a work in progress, but I have already used it for this use case multiple times : https://github.com/Gepetto/flakoboros

wentasah:
Interesting. It looks similar to [nix-ros-workspace](https://github.com/hacker1024/nix-ros-workspace), but flake-based.

PR #864, opened 2026-06-04
modules/ros2: Add applyOverlay option, default to false

znaniye:
The module unconditionally injects `nix-ros-overlay` into `nixpkgs.overlays`,
which double-applies the overlay when the consumer already added
`nix-ros-overlay.overlays.default` to their `pkgs` instantiation (e.g. via
`flake-parts`, `blueprint`, or a manual `import nixpkgs`). Double application
duplicates package patches and breaks builds (e.g. `libfyaml`
reversed-patch errors, that was my case).

Add a `services.ros2.applyOverlay` option (default `false`) that controls
whether the module injects the overlay itself. Consumers are now
expected to add `nix-ros-overlay.overlays.default` to `nixpkgs.overlays`
explicitly, which removes the ambiguity about where the overlay is
applied. Setting `applyOverlay = true` restores the previous behavior for
users who prefer the module to manage the overlay.

An assertion catches the misconfiguration where `applyOverlay` is `false`
but `pkgs.rosPackages` is missing, and a warning is emitted when
`applyOverlay` is `true` but the overlay is already present in `pkgs`.

This is a breaking change: existing configurations that rely on the
module injecting the overlay must either set `applyOverlay = true` or
add the overlay to `nixpkgs.overlays` themselves.

### DIFF
--- a/modules/ros2/ros.nix
+++ b/modules/ros2/ros.nix
@@ -21,6 +21,26 @@ in {
   options.services.ros2 = {
     enable = mkEnableOption "Robot Operating System, version 2";
 
+    applyOverlay = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether this module should inject `nix-ros-overlay` into
+        <option>nixpkgs.overlays</option>.
+
+        Defaults to `false`: the consumer is expected to add
+        `nix-ros-overlay.overlays.default` to their nixpkgs instantiation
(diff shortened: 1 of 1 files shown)


otavio (approved):

wentasah:
This makes sense. Normally, I'd be trying to avoid breaking changes, but my impression is that not many users use NixOS modules, so the impact will be limited. And you added the warning and assertion so impacted users will be notified with a user-friendly message. Thank you!

Issue #865, opened 2026-06-05
buildEnv: adding non-ros/python executables


nim65s:
Hi,

Our buildEnv only put ROS and Python derivations in `paths`.

Is there a way to add an executable ? Currently I would need `pkgs.v4l-utils`.

There are many workarounds that I can see, but I'm wondering if I'm missing a proper solution.

wentasah:
It seems `buildEnv` is somehow broken in `develop`. Adding non-ROS packages worked normally in the past and I believe that it works in `master`.

I'm a bit confused right now, but it seems that the current `buildEnv` fails to set `AMENT_PREFIX_PATH`, which causes a lot of problems and this is probably somehow related to your problem. I'm investigating...

nim65s:
Oh, yes, maybe https://github.com/lopsided98/nix-ros-overlay/pull/860/changes `propagatedBuildInputs = propagatedPaths.otherPackages;` dropped `super.propagatedBuildInputs`

wentasah:
#866 may fix that. Can you try it?

Edit: It sets AMENT_PREFIX_PATH in nix-ros-overlay examples, but in my project, it is still empty :-(.

wentasah:
I've updated #866 and now `buildEnv` seems to work as before.

wentasah:
I think this was resolved by #866. Feel free to reopen if there are still some issues with buildEnv.

PR #866, opened 2026-06-05
Fix buildEnv

wentasah:
Recent commit 228044c0e9 ("buildEnv: update for https://github.com/NixOS/nixpkgs/pull/434815", 2026-05-27) forgot to
run `fixupPhase`, which was run by the previous implementation. This, among other things, sets `AMENT_PREFIX_PATH` variable, which is needed by many ROS tools.

This mimics the structure of the implementation before 228044c0e9 and as such should be more backward compatible. This way, if users override the `fixupPhase` or set {pre,post}{Build,Fixup} hooks, the result should behave as before 228044c0e9.

### DIFF
--- a/distros/build-env/default.nix
+++ b/distros/build-env/default.nix
@@ -112,5 +112,19 @@ let
         '';
       };
     };
-  }));
+  })).overrideAttrs ({ buildCommand, ...}: {
+    # Hack to execute buildPhase and fixupPhase instead of just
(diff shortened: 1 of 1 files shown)


nim65s:
ah, yes, that looks very plausible, thanks for that !
I tried to test this, but this is not conclusive for now, I'm probably too confused, I'll try again later

wentasah:
This clearly fixes problems I have so I'm merging it. If something is still missing, let me know in #865.

PR #878, opened 2026-06-15
mavlink: Don't depend on python3Packages.future

wentasah:
This gets rid of the following error:

    error: future-1.0.0 not supported for interpreter python3.13

This is a better approach than #877.

### DIFF
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -125,13 +125,18 @@ let
     });
 
     mavlink = rosSuper.mavlink.overrideAttrs ({
-      postPatch ? "", ...
+      postPatch ? "", buildInputs ? [], ...
     }: {
       postPatch = postPatch + ''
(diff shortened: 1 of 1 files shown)


klaxalk:
@wentasah the author's of Mavlink say the dependency needs to stay for backward compatibility with LTS releases: https://github.com/ros2-gbp/mavlink-gbp-release/issues/1#issuecomment-4765619836

wentasah:
Thanks for the info. BTW, do you use `pymavlink` in your workflow? Or, in other words, if somebody uses it, is it broken because it needs the `future` module, which we now patch out from the dependencies?

klaxalk:
`pymavlink` is used to generate mavlink message headers from template xml definitions. However, it seems to work ok (the compiled package contains ok headers and the downstream packages are happy). I am not sure about the internals, but it seems to me that they actually don't need `future` for Humble+. So I don't think we have a problem :-).

PR #883, opened 2026-06-21
{jazzy,kilted}.gz-launch-vendor: fix compatibility with protobof v30

nim65s:
ref. https://github.com/gazebosim/gz-launch/pull/319


wentasah:
Thanks!

PR #904, opened 2026-08-07
Convert buildEnv to extendMkDerivation

wentasah:
This allows overriding parameters like `underlay` or `wrapPrograms` via overrideAttrs.

This commit introduces minimal changes to have a readable diff. In the next commit, we will indent and format the expressions.


wentasah:
Rebased to current `develop`.

nim65s (approved):
looks better, indeed, thanks !

And I can confirm that my workaround fiddling about `PKG_CONFIG` is still working with that instead of #902

PR #905, opened 2026-08-07
Add Qt5/6-independent package set alias

wentasah:
More and more ROS packages depend on rosdep entries that are mapped to either Qt5 or Qt6 version of a package depending on used Ubuntu version. To implement the same in Nix, we introduce package set qt5or6, which is mapped to different Qt versions based on used ROS distributions. Rosdep entries are updated to use this package set.

### DIFF
--- a/distros/distro-overlay.nix
+++ b/distros/distro-overlay.nix
@@ -1,7 +1,7 @@
 { version, distro }:
 self: super:
 let
-  useQt5 = builtins.elem distro ["humble" "jazzy" "kilted"];
+  qtVersion = if (builtins.elem distro ["humble" "jazzy" "kilted"]) then "5" else "6";
 
   pythonOverridesFor = with self.lib; prevPython: prevPython // {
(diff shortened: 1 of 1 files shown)


nim65s (approved):
I like this a lot, thanks !

distros/distro-overlay.nix
```diff
@@ -1,6 +1,8 @@
 { version, distro }:
 self: super:
 let
+  useQt5 = builtins.elem distro ["humble" "jazzy" "kilted"];
```
nim65s:
```suggestion
  qtVersion = if (builtins.elem distro ["humble" "jazzy" "kilted"]) then "5" else "6";
```

Maybe with a string instead of a bool we could facilitate downstream retrieval with `qtVersion = lib.versions.major qt5or6.qtbase.version`

That is also more future proof for qt7 x)

wentasah: Why not. Thanks.

PR #910, opened 2026-08-13
foxglove-bridge: add Darwin SDK platforms and hashes

matias-albacore:
`foxglove-bridge` fails to **evaluate** on Darwin:

```
error: attribute 'aarch64-darwin' missing
at distros/jazzy/overrides.nix:295:16:
   295|         hash = systemToHash.${self.stdenv.hostPlatform.system};
```

`systemToPlatform` and `systemToHash` map only `x86_64-linux` and `aarch64-linux`, and both are indexed with a bare `.${system}`, so any other system is an eval error rather than a graceful "unsupported".

Foxglove publishes Darwin C++ SDK archives for the same `0.26.0` release the expression already pins:

```
foxglove-v0.26.0-cpp-aarch64-apple-darwin.zip
foxglove-v0.26.0-cpp-x86_64-apple-darwin.zip
```

so both entries just need mapping. Hashes obtained with `nix store prefetch-file` against the exact URLs the expression constructs.

The `foxglove-bridge` block is duplicated verbatim across `humble`, `jazzy`, `kilted`, `lyrical` and `rolling` — same SDK version, same hashes — so all five are updated together. Happy to hoist it into `distros/ros2-overlay.nix` instead if you would prefer that; it seemed out of scope for a fix.

## Verified

Evaluated against nixpkgs-weekly with this overlay only:

- `aarch64-darwin`: `rosPackages.jazzy.foxglove-bridge` evaluates. Fails on master.
- `x86_64-linux`: unchanged, still evaluates.

Eval-level only on Darwin — CI here covers `x86_64-linux` and `aarch64-linux`, and I have no Darwin builder. No effect on Linux either way.

### DIFF
--- a/distros/humble/overrides.nix
+++ b/distros/humble/overrides.nix
@@ -290,10 +290,14 @@ in with lib; {
       systemToPlatform = {
         "x86_64-linux" = "x86_64-unknown-linux-gnu";
         "aarch64-linux" = "aarch64-unknown-linux-gnu";
+        "x86_64-darwin" = "x86_64-apple-darwin";
+        "aarch64-darwin" = "aarch64-apple-darwin";
       };
       systemToHash = {
         "x86_64-linux" = "sha256-j7A7BhEneIrXCJGNGG101P21hxixHP+XydYFUTq1fMY=";
         "aarch64-linux" = "sha256-BxDy7Fq8OVSs9iA7k6lnaNzNGddCr5t4/Nj7u6X2Ilo=";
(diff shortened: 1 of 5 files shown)


matias-albacore:
Closing this — we're carrying the fix downstream in our own overlay for now rather than asking you to take on Darwin support that CI here doesn't cover. Sorry for the noise.

wentasah:
I think we can include this. It's not a big deal to have the hashes updated even without CI. At least this gave me enough motivation to prepare #911, which simplifies updates also for other platforms :-)

wentasah:
Since the original branch is deleted, I'll include this in #911.

Issue #754, opened 2025-11-29
Cachix cache issues (too small capcity)


wentasah:
In recent weeks I spent a lot of time with investigating and fixing problems with our binary cache. This is a summary of what I found and a start of discussion about future steps.

The biggest problem is that [our cache](https://app.cachix.org/cache/ros) has limited size.

Currently, binary packages for each ROS distribution consume about 5 GB (see [logs from my tests](https://github.com/wentasah/nix-ros-overlay/actions/runs/19786847273/job/56694507059#step:6:6773)). In addition to that, each package also produces debug info in a separate output, which also includes package source code and this needs roughly additional 5 GB. We support four distributions (humble, jazzy, kilted and rolling) and two architectures (x86 and ARM) and this is a lot of data in total.

Cachix compresses the data a bit so in my experiments, building all the above combinations resulted in 54 GB in the cache.

I also added Cachix pinning, to always keep one version of the binary package in the cache. I pin only the `$out` outputs, not the `$debug` ones. If Cachix statistics do not lie, the pinned paths consume **8.2 GB** of the cache. I don't know what's the size of the cache, but my personal free cache has capacity of 10 GB, so 8.2 GB looks good. The problem is that this is just for the `develop` branch and in the ideal case, we would want to pin the `master` branch too, which will likely result in more than 10 GB.

@lopsided98 I don't have access to the Cachix `ros` cache. Can you verify what's the current capacity? I'd also need to configure secrets in the GitHub repo for pinning to work. See [this error](https://github.com/lopsided98/nix-ros-overlay/actions/runs/19795776271/job/56715772191#step:6:3812) (`export CACHIX_AUTH_TOKEN=<token...>`). I don't understand how CI can push to the cache, and fail to pin. In tests with my cache, pinning worked well.

If the above error with pinning is fixed, we will see how the cache behaves with it. But I'm afraid that 10 GB capacity would not be sufficient. Then, we can try the following:

- Ask Cachix to kindly increase our cache quota.
- Switch to nix-community cache, which IFAIK already has increased quota. This would mean that all users would have to change their cache configuration.
- Switch to [our university cache](https://attic.iid.ciirc.cvut.cz/ros). Users would have to change their configuration too and there would be a single point of failure – me. Additionally, I'm not sure how good connectivity it would have in different parts of the world.

  See also #753 which proposes using this cache just for CI.

If you have any other idea or comment, feel free to let me know.

otavio:
[...]

PR #913, opened 2026-08-14
Fix some eval and build errors on Darwin

### LINKED ISSUE #863: MacOS support
There's been a couple attempts to support MacOS in the past. I'm looking into trying it again over the next few months. A friend has recently offered to loan me his M-series Macbook so I can build packages natively. From initial attempts on my own machine (x86_64 NixOS) using the --system aarch64-darwin flag, it seems like some changes may be needed for rosdistro and ros2nix (Modifying linux-specific keys to add a macos nix package) to avoid overriding every package that depends on these packages - although this is where I'll start.

I won't be testing on any x86_64 MacOS machines since the 26.05 release announced that x86_64-darwin is being deprecated (https://nixos.org/blog/announcements/2026/nixos-2605/).

I was also thinking of adding a MacOS runner to the GitHub actions. @wentasah I don't suppose it would be possible to add MacOS to your hydra instance as well?

wentasah:
This is a continuation of #909 proposed by @matias-albacore. Since we're dealing with eval errors, it's not necessary to have full Darwin support in CI. I've added a simple CI job, that evalues `ros-base` and `desktop` packages for Darwin platforms. Without this PR the script failed, now, it succeeds. Hopefully, it will be sufficient for discovering biggest regressions. For build errors, I added a simple CI (GitHub workflow), which builds `ros-base` for every distro.

Tagging users who have dealt with this overlay and Darwin recently: @BrockoliniMorgan (#863), @klaxalk.

Closes #863.

### DIFF
--- a/distros/kilted/overrides.nix
+++ b/distros/kilted/overrides.nix
@@ -83,6 +83,20 @@ in {
     '';
   });
 
+  fastdds = rosSuper.fastdds.overrideAttrs ({
+    postPatch ? "", ...
+  }: lib.optionalAttrs self.stdenv.isDarwin {
+    # `value` (std::string) is implicitly constructed from a null
+    # `const char*` for the false branch, tripping -Wnonnull on
+    # Apple Clang. Fixed upstream in Fast-DDS 3.3 by
+    # https://github.com/eProsima/Fast-DDS/pull/6415.
+    postPatch = postPatch + ''
+      substituteInPlace src/cpp/fastdds/xtypes/type_representation/TypeObjectRegistry.cpp --replace-fail \
+        'TypeObjectUtils::build_annotation_parameter_value(!value.empty() ? value : 0);' \
+        "TypeObjectUtils::build_annotation_parameter_value(!value.empty() ? value[0] : '\\0');"
+    '';
+  });
+
   foxglove-bridge = rosSuper.foxglove-bridge.overrideAttrs({
     postPatch ? "", ...
   }: {
--- a/distros/kilted/overrides.nix
+++ b/distros/kilted/overrides.nix
@@ -85,7 +85,7 @@ in {
 
   fastdds = rosSuper.fastdds.overrideAttrs ({
     postPatch ? "", ...
-  }: lib.optionalAttrs self.stdenv.isDarwin {
+  }: lib.optionalAttrs self.stdenv.hostPlatform.isDarwin {
     # `value` (std::string) is implicitly constructed from a null
     # `const char*` for the false branch, tripping -Wnonnull on
     # Apple Clang. Fixed upstream in Fast-DDS 3.3 by
--- a/maintainers/scripts/eval-darwin.sh
+++ b/maintainers/scripts/eval-darwin.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -i bash --pure  -I nixpkgs=. --packages nix findutils
+# shellcheck shell=bash
+set -euo pipefail
+
+system=${1:-aarch64-darwin}
+
+for distro in $(find distros -mindepth 1 -maxdepth 1 -type d -not -name '*-*' -printf '%f\n'); do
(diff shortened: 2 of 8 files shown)


.github/workflows/eval-darwin.yaml
```diff
@@ -0,0 +1,12 @@
+name: "Eval Darwin"
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+jobs:
+  eval-darwin:
+    runs-on: ubuntu-latest
```
nim65s: this could be a generic eval job running on a matrix with `ubuntu-latest`, `ubuntu-24.04-arm` and `macos-latest`

maintainers/scripts/eval-darwin.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+system=aarch64-darwin
```
nim65s: and here a conditional `$1`

.github/workflows/eval-darwin.yaml
wentasah: Do you think there will be a difference if you run eval on different platforms?

wentasah: But I didn't notice that github provides `macos-latest` runners. So we may actually run the build on MacOS too, no?

nim65s: No, I don't think the eval will be different. 
We can run builds, but I don't think we can cache the results, unless you want to give github a push access to your attic

wentasah:
With this, `ros-base` in all distros builds on `aarch64-darwin` without errors and we also have CI for it.

distros/kilted/overrides.nix
```diff
@@ -83,11 +83,19 @@ in {
...
   });
 
-  foonathan-memory-vendor = lib.patchExternalProjectGit rosSuper.foonathan-memory-vendor {
-    url = "https://github.com/foonathan/memory.git";
-    rev = "v0.7-3";
-    fetchgitArgs.hash = "sha256-nLBnxPbPKiLCFF2TJgD/eJKJJfzktVBW3SRW2m3WK/s=";
-  };
+  fastdds = rosSuper.fastdds.overrideAttrs ({
+    postPatch ? "", ...
+  }: lib.optionalAttrs self.stdenv.isDarwin {
```
nim65s: in nixpkgs we use `stdenv.hostPlatform.isDarwin` instead of `stdenv.isDarwin`

nim65s (commented):
super cool, thanks !
And yes, another cache make sense for another OS :)
Should we add this `ros-darwin.cachix.org` in the README ?

wentasah:
> Should we add this `ros-darwin.cachix.org` in the README ?

Let's wait for https://github.com/lopsided98/nix-ros-overlay/issues/754#issuecomment-5295611174. And even then, I'm not sure how much I want to support MacOS. ros-base has not many packages so the cache is not much useful now. It seems there will be a lot of build failures in other packages and debugging them by pushing to Github for CI is not very convenient.

distros/kilted/overrides.nix
wentasah: Right, will fix it.

klaxalk:
Hello all, this is definitely appreciated! We are now running a drone with MAC OS and we are dealing with the same issues. If anybody is interested, I have a **very raw** custom ROS2 Jazzy overlay with our own packages on top of `nix-ros-overlay` that is built and cashed for amd64, arm64 and darwin. Maybe it might be useful to some of you.
* https://github.com/ctu-mrs/nix-mrs-overlay/

I am now paying for a cachix cache, just because it makes my live so much easier. I am looking into hosting our own soon though.
* https://app.cachix.org/cache/ctu-mrs

Any feedback and collaboration is welcomed :-).

wentasah:
> Maybe it might be useful to some of you.
> 
>     * https://github.com/ctu-mrs/nix-mrs-overlay/

Yes, I was looking there when working on this PR :-)

PR #914, opened 2026-08-14
zenoh-cpp-vendor: USE_SYSTEM_ZENOH

nim65s:


wentasah:
Do you think this is safe? I'm not sure how tight is dependency between rmw_zenoh and the actual zenoh version, but I'd expect there might be subtle incompatibilities that would break things if rmw_zenoh is used with different zenoh version. Do you have experience with that?

nim65s:
I believe that would be the job of zenoh-cpp-vendor upstream to manage versions constraints since they are the one exposing `USE_SYSTEM_ZENOH`, and that would be trivial to express in CMake given the same devellopers are providing eg. `lib/cmake/zenohcxx/zenohcxxConfigVersion.cmake` in one package and consuming it in another.

I also believe `rmw_zenoh` tests should catch anything wrong… But I forgot we don't run those, I'll double check this.

nim65s:
```
$ for d in humble jazzy kilted lyrical rolling 
           nix log .#$d.test-rmw-zenoh-cpp | grep 'test_rmw_zenoh_session .'
       end
7/7 Test #1: test_rmw_zenoh_session ...........   Passed    0.77 sec
7/7 Test #1: test_rmw_zenoh_session ...........   Passed    0.78 sec
7/7 Test #1: test_rmw_zenoh_session ...........   Passed    0.80 sec
7/7 Test #1: test_rmw_zenoh_session ...........   Passed    0.79 sec
7/7 Test #1: test_rmw_zenoh_session ...........   Passed    0.81 sec
```

Yes, so now I'm pretty sure that we can use zenoh-cpp 1.9.0 on humble jazzy kilted lyrical and rolling :)

Maybe we don't need to keep the tests though, I'll let you decide

nim65s:
btw, the test-rmw-zenoh-cpp packages have been generated from
```toml
[humble."ros2/rmw_zenoh"]
branch = "humble"
[jazzy."ros2/rmw_zenoh"]
branch = "jazzy"
[kilted."ros2/rmw_zenoh"]
branch = "kilted"
[lyrical."ros2/rmw_zenoh"]
branch = "lyrical"
[rolling."ros2/rmw_zenoh"]
branch = "rolling"

wentasah:
I'll have look at this later, perhaps during weekend.

